PR #330, opened 2024-11-20
chore: refactor static setConnectionProvider method

sophia-bq:
### Summary

Refactor setting a non-default ConnectionProvider to be done through a connection property instead of a static method. 

### Description

Adds a new connection property `connectionProvider` that allows users to pass a non-default provider through a connection property. 

This replaces the existing static `setConnectionProvider` method that gives any code access to modify the connection provider manager on the go. 

The value of `ConnectionProviderManager.connProvider` which is updated by the connection property remains static.

### DIFF
--- a/common/lib/aws_client.ts
+++ b/common/lib/aws_client.ts
@@ -64,7 +64,7 @@ export abstract class AwsClient extends EventEmitter {
     this.pluginManager = new PluginManager(
       container,
       this.properties,
-      new ConnectionProviderManager(new DriverConnectionProvider(), null, WrapperProperties.CONNECTION_PROVIDER.get(this.properties)),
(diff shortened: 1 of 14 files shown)


common/lib/aws_client.ts
```diff
@@ -63,7 +64,7 @@ export abstract class AwsClient extends EventEmitter {
     this.pluginManager = new PluginManager(
       container,
       this.properties,
-      new ConnectionProviderManager(new DriverConnectionProvider(), null),
+      new ConnectionProviderManager(new DriverConnectionProvider(), null, WrapperProperties.CONNECTION_PROVIDER.get(this.properties)),
```
sergiyvamz, line 67: Why not to use effectiveProvider in stead of adding a new parameter?

sergiyvamz (approved):

PR #332, opened 2024-11-22
chore: optimize getHostInfoByStrategy in PluginManager

sophia-bq:
### Summary

Cache the getHostInfoByStrategy chain like the execute and connect pipelines so the plugin chain is not reconstructed every call.

### DIFF
--- a/common/lib/plugin_manager.ts
+++ b/common/lib/plugin_manager.ts
@@ -63,6 +63,7 @@ class PluginChain<T> {
 
 export class PluginManager {
   private static readonly PLUGIN_CHAIN_CACHE = new Map<[string, HostInfo], PluginChain<any>>();
+  private static readonly STRATEGY_PLUGIN_CHAIN_CACHE = new Map<ConnectionPlugin[], Set<ConnectionPlugin>>();
   private static readonly ALL_METHODS: string = "*";
   private static readonly CONNECT_METHOD = "connect";
   private static readonly FORCE_CONNECT_METHOD = "forceConnect";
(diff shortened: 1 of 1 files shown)


common/lib/plugin_manager.ts
```diff
@@ -282,20 +283,26 @@ export class PluginManager {
   }
 
   getHostInfoByStrategy(role: HostRole, strategy: string, hosts?: HostInfo[]): HostInfo {
```
karenc-bq, line 285: Can you make the same optimization for acceptsStrategy since the code is very similar?

common/lib/plugin_manager.ts
```diff
@@ -282,20 +283,26 @@ export class PluginManager {
...
-
-      if (isSubscribed) {
-        try {
-          const host = plugin.getHostInfoByStrategy(role, strategy, hosts);
-          if (host) {
-            return host;
-          }
-        } catch (error) {
-          // This plugin does not support the provided strategy, ignore the exception and move on
+    if (PluginManager.GET_HOST_INFO_BY_STRATEGY_PLUGIN_CHAIN_CACHE.length === 0) {
```
karenc-bq, line 286: Could you please use `PLUGIN_CHAIN_CACHE`'s implementation as a reference?
Different connections may be using different list of plugins, your current implementation will throw an error.
See this simple test:
```ts
    const client = new AwsPGClient({
      user: PG_DB_USER,
      host: PG_DB_HOST,
      database: PG_DB_NAME,
      password: PG_DB_PASSWORD,
      port: 5432,
      plugins: "readWriteSplitting",
      readerHostSelectorStrategy: "leastConnections",
      connectionProvider: new InternalPooledConnectionProvider()
    });

    await client.connect();

    const client2 = new AwsPGClient({
      user: PG_DB_USER,
      host: PG_DB_HOST,
      database: PG_DB_NAME,
      password: PG_DB_PASSWORD,
      port: 5432,
      plugins: ""
    });

    await client2.connect();
```

karenc-bq: Also this implementation will iterate over the plugin chain twice on the initial connection. This can be optimized.

common/lib/plugin_manager.ts
```diff
@@ -63,6 +63,7 @@ class PluginChain<T> {
 
 export class PluginManager {
   private static readonly PLUGIN_CHAIN_CACHE = new Map<[string, HostInfo], PluginChain<any>>();
+  private static readonly GET_HOST_INFO_BY_STRATEGY_PLUGIN_CHAIN_CACHE = new Map<[string, PluginManager], Set<ConnectionPlugin>>();
```
karenc-bq, line 66: Why is the key `[string, PluginManager]`?
Looking at the way this map is accessed, it gets and sets with `[PluginManager.GET_HOST_INFO_BY_STRATEGY_METHOD, this]`, why do we need `PluginManager.GET_HOST_INFO_BY_STRATEGY_METHOD` as part of the key if it never changes?

karenc-bq: Will `new Map<PluginManager, Set<ConnectionPlugin>>();` work?

sophia-bq: If we want to use the same map for AcceptsStrategy yes, if we want different ones no

karenc-bq: Is it possible to have scenarios where acceptsStrategy and and getHostInfoByStrategy use different plugin chains?
Also if we want to use the same map for both lets rename the variable to `getHostInfoByStrategy`

karenc-bq: On second thought, if it is possible to have scenarios where  acceptsStrategy and getHostInfoByStrategy use different plugin chains, why do we want to use the same cache for both of them instead of just having 2 different caches?

common/lib/plugin_manager.ts
```diff
@@ -140,6 +141,7 @@ export class PluginManager {
   }
 
   async connect(hostInfo: HostInfo | null, props: Map<string, any>, isInitialConnection: boolean): Promise<ClientWrapper> {
+    const temp = this._plugins;
```
karenc-bq, line 144:
```suggestion
```

karenc-bq (approved):

PR #333, opened 2024-11-22
chore: make AwsClient#releaseResouces static

sophia-bq:
### Summary

Update releaseResources to be a static method. 

### Description

Introduced a static set to track all the plugins in PluginManager that releases all resources when Client.end is called.


joyc-bq (approved):

mysql/lib/client.ts
```diff
@@ -212,7 +213,7 @@ export class AwsMySQLClient extends AwsClient {
       },
       null
     );
-    await this.releaseResources();
+    await PluginManager.releaseResources();
```
karenc-bq, line 216: `PluginManager.releaseResources()` is made static because it is meant to be called at the end of the application. It is not supposed be called every time we close a connection. Can you please verify if we are supposed to call `this.releaseResources` here.

karenc-bq (approved):

PR #339, opened 2024-11-27
feat: wrapper client setKeepAlive method

crystall-bitquill:
### Summary

Wrapper client setKeepAlive method

### Description

- adds a setKeepAlive method to the wrapper clients, which if possible, will set keepalive based on the underlying driver


joyc-bq (approved):

pg/lib/client.ts
```diff
@@ -214,4 +214,22 @@ export class AwsPGClient extends AwsClient {
...
+      for (const [key, value] of options) {
+        this.properties.set(key, value);
+      }
+    } else {
+      for (const property of this.pluginService.getDriverDialect().keepAlivePropertyNames) {
+        this.properties.delete(property);
+      }
+    }
+
+    await this.connect();
```
karenc-bq, line 233: this method is confusing, why does `setKeepAlive` abort or create connections?

common/lib/aws_client.ts
```diff
@@ -149,4 +149,6 @@ export abstract class AwsClient extends EventEmitter {
   getPluginInstance<T>(iface: any): T {
     return this.pluginManager.getPluginInstance(iface);
   }
+
+  abstract setKeepAlive(keepAlive: boolean, options?: any): Promise<void>;
```
karenc-bq, line 153: Why is setKeepAlive a method on the `awsClient` instead of a connection property like the network timeouts?

mysql/lib/dialect/mysql2_driver_dialect.ts
```diff
@@ -26,6 +26,7 @@ import { HostInfo } from "../../../common/lib/host_info";
 
 export class MySQL2DriverDialect implements DriverDialect {
   protected dialectName: string = this.constructor.name;
+  keepAlivePropertyNames: string[] = [];
```
karenc-bq, line 29: Is there a reason this is not a static constant? can this be modified?

sergiyvamz: Why the type is string[], not just string?

crystall-bitquill: It's a string[] because there could be multiple properties, for example PG having both keepalive and keepaliveInitialDelayMillis

pg/lib/dialect/node_postgres_driver_dialect.ts
```diff
@@ -55,4 +56,14 @@ export class NodePostgresDriverDialect implements DriverDialect {
   getAwsPoolClient(props: pkgPg.PoolConfig): AwsPoolClient {
     return new AwsPgPoolClient(props);
   }
+
+  setupInitialProperties(props: Map<string, any>) {
+    if (props.has(WrapperProperties.KEEPALIVE_PROPERTIES.name)) {
```
sergiyvamz, line 61: That looks a bit overengineered. There're just two properties to setup. There's almost 0 chance that these properties change any time soon.

mysql/lib/dialect/mysql2_driver_dialect.ts
```diff
@@ -52,4 +52,6 @@ export class MySQL2DriverDialect implements DriverDialect {
   getAwsPoolClient(props: PoolOptions): AwsPoolClient {
     return new AwsMysqlPoolClient(props);
   }
+
+  setupInitialProperties(props: Map<string, any>) {}
```
sergiyvamz, line 56: Should we notify user that this feature isn't supported by mysql driver if the user provides tcpKeepAlive configuration?

common/lib/wrapper_property.ts
```diff
@@ -329,6 +329,12 @@ export class WrapperProperties {
     600_000 // 10 min
   );
 
+  static readonly KEEPALIVE_PROPERTIES = new WrapperProperty<any>(
```
sergiyvamz, line 332: Can we be more specific in terms of what type to expect (rather than `any`)?

common/lib/aws_client.ts
```diff
@@ -58,6 +58,8 @@ export abstract class AwsClient extends EventEmitter {
 
     this.properties = new Map<string, any>(Object.entries(config));
 
+    driverDialect.setupInitialProperties(this.properties);
```
sergiyvamz, line 61: I think it's better to setup these properties in connection provider connect() method.

sergiyvamz: or in driver dialect connect()

mysql/lib/dialect/mysql2_driver_dialect.ts
crystall-bitquill: I've added a note in the UsingTheNodejsWrapper.md file, do you think that's sufficient?

karenc-bq (approved):

mysql/lib/dialect/mysql2_driver_dialect.ts
```diff
@@ -52,4 +52,6 @@ export class MySQL2DriverDialect implements DriverDialect {
   getAwsPoolClient(props: PoolOptions): AwsPoolClient {
     return new AwsMysqlPoolClient(props);
   }
+
+  setKeepAliveProperties(props: Map<string, any>, keepAliveProps: any) {}
```
sergiyvamz, line 56: Since mysql driver doesn't support tcpKeepAlive I'd suggest to check keepAliveProp and if it set then raise an exception saying that the driver doesn't support it.

sergiyvamz (approved):

PR #342, opened 2024-11-29
feat: wrapper connect and query timeouts

crystall-bitquill:


mysql/lib/dialect/mysql2_driver_dialect.ts
```diff
@@ -52,4 +54,12 @@ export class MySQL2DriverDialect implements DriverDialect {
   getAwsPoolClient(props: PoolOptions): AwsPoolClient {
     return new AwsMysqlPoolClient(props);
   }
+
+  setupDriverProperties(props: Map<string, any>) {
```
karenc-bq, line 58: Can we use a more specific method name.
This method is setting up network related timeouts.

common/lib/wrapper_property.ts
```diff
@@ -77,6 +77,18 @@ export class WrapperProperties {
     20000
   );
 
+  static readonly WRAPPER_CONNECT_TIMEOUT = new WrapperProperty<number>(
```
karenc-bq, line 80: We need to mark MySQLQueryTimeout as deprecated in here and in docs

mysql/lib/dialect/mysql2_driver_dialect.ts
karenc-bq: What do you think of `setNetworkTimeouts`

common/lib/aws_client.ts
```diff
@@ -58,6 +58,8 @@ export abstract class AwsClient extends EventEmitter {
 
     this.properties = new Map<string, any>(Object.entries(config));
 
+    driverDialect.setupDriverProperties(this.properties);
```
karenc-bq, line 61: Is there a reason this is called so early before we actually use it?
Do we really need this extra public facing method? 
What do you think of setting up the connect timeout directly  within `DriverDialect#connect`
```
async connect(hostInfo: HostInfo, props: Map<string, any>): Promise<ClientWrapper> {
    const wrapperConnectTimeout = props.get(WrapperProperties.WRAPPER_CONNECT_TIMEOUT.name);

    if (wrapperConnectTimeout) {
      props.set(this.connectTimeoutPropertyName, wrapperConnectTimeout);
    }
    const targetClient = await createConnection(WrapperProperties.removeWrapperProperties(props));
    return Promise.resolve(new MySQLClientWrapper(targetClient, hostInfo, props, this));
  }
```

common/lib/mysql_client_wrapper.ts
```diff
@@ -34,15 +37,21 @@ export class MySQLClientWrapper implements ClientWrapper {
...
+  constructor(targetClient: any, hostInfo: HostInfo, properties: Map<string, any>, driverDialect: DriverDialect) {
     this.client = targetClient;
     this.hostInfo = hostInfo;
     this.properties = properties;
+    this.driverDialect = driverDialect;
     this.id = uniqueId("MySQLClient_");
   }
 
   query(sql: any): Promise<any> {
+    const queryTimeout = this.properties.get(WrapperProperties.WRAPPER_QUERY_TIMEOUT.name);
```
karenc-bq, line 51: For MySQL we are already using our custom timeout wrapper. What is the purpose of adding this timeout here as well?

common/lib/mysql_client_wrapper.ts
```diff
@@ -34,15 +37,21 @@ export class MySQLClientWrapper implements ClientWrapper {
...
+    if (queryTimeout && !sql[this.driverDialect.queryTimeoutPropertyName]) {
+      sql[this.driverDialect.queryTimeoutPropertyName] = queryTimeout;
```
karenc-bq, line 53: This line is very driver specific, it assumes the sql is a obj instead of a string so I think this logic belongs in driver dialect as well.
What do you think of adding a static method
```ts
  static prepareQuery(sql: any, properties: Map<string, any>) {
    const queryTimeout = properties.get(WrapperProperties.WRAPPER_QUERY_TIMEOUT.name);
    if (queryTimeout && !sql[MySQL2DriverDialect.queryTimeoutPropertyName]) {
      sql[MySQL2DriverDialect.queryTimeoutPropertyName] = queryTimeout;
    }
    return sql;
  }
```

This way this method can be simplified and you won't need to pass driver dialects around anymore.
```
  query(sql: any): Promise<any> {
    return this.client?.query(DriverDialect.prepareQuery(sql, this.properties));
  }
```

pg/lib/dialect/node_postgres_driver_dialect.ts
```diff
@@ -27,6 +27,8 @@ import { PgClientWrapper } from "../../../common/lib/pg_client_wrapper";
 import { HostInfo } from "../../../common/lib/host_info";
 
 export class NodePostgresDriverDialect implements DriverDialect {
+  readonly connectTimeoutPropertyName = "connectionTimeoutMillis";
+  readonly queryTimeoutPropertyName = "query_timeout";
```
karenc-bq, line 31: can be static

common/lib/aws_client.ts
karenc-bq: On second thought, I am okay with keeping the code contained as a method if it looks cleaner.
I think we should separate them into individual methods like setConnectTimeout and setQueryTimeout, and call them before we connect or query, so the intent is more clear.

karenc-bq: What do you think @sergiyvamz

common/lib/mysql_client_wrapper.ts
karenc-bq: @sergiyvamz

sergiyvamz: I partially agree here. I think this piece of code is very driver specific and it should be moved to driver dialect class. And I don't agree with @karenc-bq about static method. That makes a static bond between client and dialect. It'd cause issues mocking and testing. Why not to make a `getQueryTimeout()` method in driver dialect class? The method can do type conversion if needed.

tests/integration/container/tests/utils/driver_helper.ts
```diff
@@ -117,8 +117,8 @@ export class DriverHelper {
       props["monitoring_query_timeout"] = 3000;
     } else if (engine === DatabaseEngine.MYSQL && performance) {
       props["connectTimeout"] = 3000;
-      props["monitoring_mysqlQueryTimeout"] = 3000;
-      props["mysqlQueryTimeout"] = 120000;
+      props["monitoring_wrapperQueryTimeout"] = 3000;
```
sergiyvamz, line 120: It might be too late (or might be not) to consolidate our parameter names. They should use either camel syntax or underscores. It seems we have a mix of both. Probably it's out of scope of this PR.

common/lib/mysql_client_wrapper.ts
karenc-bq: If I understood you correctly, the `getQueryTimeout()` method will be used like this for mysql
```ts
return this.client?.query(this.driverDialect.getQueryTimeout(sql, this.properties));
```
and like this for pg?
```
this.properties = this.driverDialect.getQueryTimeout(sql, this.properties);
return this.client?.queryWithTimeout(sql, this.properties);
```

common/lib/mysql_client_wrapper.ts
crystall-bitquill: We have the wrapper timeout, but the base driver still accepts timeout values in the query method. Should we ignore it? Users would still be able to set that parameter through their initial query.

common/lib/wrapper_property.ts
```diff
@@ -71,12 +71,27 @@ export class WrapperProperties {
     null
   );
 
+  /**
+   * @deprecated since version 1.0.1 and replaced by wrapper property WRAPPER_QUERY_TIMEOUT.
```
karenc-bq, line 75: Should be 1.1.0.
Bumping to 1.1.0 because configuration profiles is a feature release.

docs/using-the-nodejs-wrapper/UsingTheNodejsWrapper.md
```diff
@@ -50,7 +50,7 @@ These parameters are applicable to any instance of the AWS Advanced NodeJS Wrapp
...
+| ~~`mysqlQueryTimeout`~~        | `number`  | No                                                                               | Query timeout in milliseconds. This is only applicable when using the AwsMySQLClient. To set query timeout for the AwsPGClient, please use the built-in `query_timeout` parameter. See the `node-postgres` [documentation](https://node-postgres.com/apis/client) for more details..                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          | 20000                                                                                                                                                                        |
```
karenc-bq, line 53: Lets also mention that this is deprecated and that applications should opt for wrapperQueryTimeout instead.
Also there is an extra full-stop at the end.

docs/using-the-nodejs-wrapper/UsingTheNodejsWrapper.md
```diff
@@ -50,7 +50,9 @@ These parameters are applicable to any instance of the AWS Advanced NodeJS Wrapp
...
+| ~~`mysqlQueryTimeout`~~        | `number`  | No                                                                               | This parameter has been deprecated, applications should use the `wrapperQueryTimeout` parameter instead. <br/><br/> Query timeout in milliseconds. This is only applicable when using the AwsMySQLClient. To set query timeout for the AwsPGClient, please use the built-in `query_timeout` parameter. See the `node-postgres` [documentation](https://node-postgres.com/apis/client) for more details.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       | 20000                                                                                                                                                                        |
```
karenc-bq, line 53: Need to mention the version. e.g. `has been deprecated since version....`
Can we add a `version` column so users don't try to use newer parameters on older driver versions

common/lib/aws_client.ts
crystall-bitquill: This is called here in order to set up the properties with the correct timeout values prior to creating objects like the telemetryFactory, pluginService and pluginManager, all of which accept a copy of the properties.

mysql/lib/dialect/mysql2_driver_dialect.ts
```diff
@@ -52,4 +54,27 @@ export class MySQL2DriverDialect implements DriverDialect {
...
+  }
+
+  setQueryTimeout(props: Map<string, any>, sql?: any) {
+    const queryTimeout = props.get(WrapperProperties.WRAPPER_QUERY_TIMEOUT.name);
+    if (queryTimeout && !sql[this.getQueryTimeoutPropertyName()]) {
+      sql[this.getQueryTimeoutPropertyName()] = queryTimeout;
+    }
+  }
+
+  getQueryTimeoutPropertyName(): string {
```
karenc-bq, line 77: Curious if this method is required if the property name is only used locally now.

common/lib/driver_dialect/driver_dialect.ts
```diff
@@ -27,4 +27,12 @@ export interface DriverDialect {
   preparePoolClientProperties(props: Map<string, any>, poolConfig: AwsPoolConfig | undefined): any;
 
   getAwsPoolClient(props: any): AwsPoolClient;
+
+  setConnectTimeout(props: Map<string, any>): void;
+
+  setupInitialNetworkTimeouts(props: Map<string, any>): void;
```
sergiyvamz, line 33: What means "initial"? How it's different from queryTimeout?

crystall-bitquill: This method is to set up any properties used in the client configuration. The separate set connect/query timeout methods are from a [suggestion](https://github.com/aws/aws-advanced-nodejs-wrapper/pull/342#discussion_r1870194732), they may or may not be called in `setupInitialNetworkTimeouts`.

common/lib/aws_client.ts
sergiyvamz: I don't think it's important to setup driver specific properties before creating mentioned objects. These objects (telemetry factory, plugin service, etc.) are not suppose to access such driver specific properties and they don't know about them.

sergiyvamz: I agree with Karen's idea to move this code to driver dialect connect().

common/lib/aws_client.ts
```diff
@@ -58,6 +58,8 @@ export abstract class AwsClient extends EventEmitter {
 
     this.properties = new Map<string, any>(Object.entries(config));
 
+    driverDialect.setupInitialNetworkTimeouts(this.properties);
```
sergiyvamz, line 61: Need to move to driver dialect connect()

common/lib/driver_dialect/driver_dialect.ts
karenc-bq: I am leaning towards renaming this to setNetworkTimeouts since this method can technically be called any time during the application workflow.

karenc-bq (approved):

common/lib/utils/client_utils.ts
```diff
@@ -18,15 +18,26 @@ import { getTimeoutTask } from "./utils";
...
-    const timeoutTask = getTimeoutTask(
-      timer,
-      Messages.get("ClientUtils.queryTaskTimeout"),
-      timeValue ?? WrapperProperties.INTERNAL_QUERY_TIMEOUT.get(props)
-    );
+    let timeout = timeValue;
+    if (props.has(WrapperProperties.INTERNAL_QUERY_TIMEOUT.name)) {
+      timeout = WrapperProperties.INTERNAL_QUERY_TIMEOUT.get(props);
+      if (!ClientUtils.hasWarnedDeprecation) {
+        logger.warn(
```
sergiyvamz, line 31: I'd warn a user again and again.

pg/lib/dialect/node_postgres_driver_dialect.ts
```diff
@@ -27,14 +27,19 @@ import { PgClientWrapper } from "../../../common/lib/pg_client_wrapper";
...
 
   getDialectName(): string {
     return this.dialectName;
   }
 
   async connect(hostInfo: HostInfo, props: Map<string, any>): Promise<ClientWrapper> {
-    const targetClient = new pkgPg.Client(WrapperProperties.removeWrapperProperties(props));
+    const driverProperties = WrapperProperties.removeWrapperProperties(props);
+    this.setConnectTimeout(driverProperties, props.get(WrapperProperties.WRAPPER_CONNECT_TIMEOUT.name));
+    this.setConnectTimeout(driverProperties, props.get(WrapperProperties.WRAPPER_QUERY_TIMEOUT.name));
```
sergiyvamz, line 41: this.setQueryTimeout() ?

sergiyvamz (approved):

PR #353, opened 2024-12-13
test: configure fake timers for iam tests

sophia-bq:
### Summary

Updates integration tests. 

### Description

- Configures fakeTimers to not fake 'nextTick' for mysql to validate connection successfully. 
- Resets timers after all iam authentication tests. 
- Correctly orders the release of resources when a non default provider is used.

### DIFF
--- a/common/lib/driver_connection_provider.ts
+++ b/common/lib/driver_connection_provider.ts
@@ -97,7 +97,7 @@ export class DriverConnectionProvider implements ConnectionProvider {
           JSON.stringify(Object.fromEntries(maskProperties(resultProps)))
       );
 
-      resultTargetClient = driverDialect.connect(hostInfo, resultProps);
+      resultTargetClient = await driverDialect.connect(hostInfo, resultProps);
     }
     pluginService.attachErrorListener(resultTargetClient);
     return resultTargetClient;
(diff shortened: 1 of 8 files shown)


common/lib/driver_connection_provider.ts
```diff
@@ -97,7 +97,7 @@ export class DriverConnectionProvider implements ConnectionProvider {
           JSON.stringify(Object.fromEntries(maskProperties(resultProps)))
       );
 
-      resultTargetClient = driverDialect.connect(hostInfo, resultProps);
+      resultTargetClient = await driverDialect.connect(hostInfo, resultProps);
```
karenc-bq, line 100: Is this required? I think I remembered this causing some issues for one of the drivers.

examples/aws_driver_example/aws_internal_connection_pool_password_warning_mysql_example.ts
```diff
@@ -72,6 +72,9 @@ try {
 } finally {
   await newClient.end();
 }
+// Clean up resources used by the plugins.
+await PluginManager.releaseResources();
+
 // Closes all pools and removes all cached pool connections.
 await provider.releaseResources();
```
karenc-bq, line 79: These two methods are meant to be called at the end of the application. Curious if we should move both of them to the end, after we are done with newClient2.

karenc-bq: Same comment for the aws_internal_connection_pooling_mysql_example.

karenc-bq (approved):

PR #363, opened 2025-01-03
feat: failover2

sophia-bq:


common/lib/utils/locales/en.json
```diff
@@ -193,5 +195,6 @@
   "ConfigurationProfileBuilder.notFound": "Configuration profile '%s' not found.",
   "ConfigurationProfileBuilder.profileNameRequired": "Profile name is required.",
   "ConfigurationProfileBuilder.canNotUpdateKnownPreset": "Can't add or update a built-in preset configuration profile '%s'.",
-  "AwsClient.configurationProfileNotFound": "Configuration profile '%s' not found."
+  "AwsClient.configurationProfileNotFound": "Configuration profile '%s' not found.",
+  "Failover.unableToConnect": "Failover.unableToConnect"
```
karenc-bq, line 199: Need to fix this log message.

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,482 @@
...
+import { PluginService } from "../../plugin_service";
+import { HostAvailability } from "../../host_availability/host_availability";
+import { logTopology, sleep } from "../../utils/utils";
+import { logger } from "../../../logutils";
+import { HostRole } from "../../host_role";
+import { ClientWrapper } from "../../client_wrapper";
+import { AwsWrapperError } from "../../utils/errors";
+import { MonitoringRdsHostListProvider } from "./monitoring_host_list_provider";
+
+export interface ClusterToplogyMonitor {
```
karenc-bq, line 28: Typo in topology

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,482 @@
...
+    this.releaseTopologyUpdate = () => {
+      done();
+    };
+  });
+
+  constructor(
+    clusterId: string,
+    topologyMap: CacheMap<string, HostInfo[]>,
+    initialHostSpec: HostInfo,
+    props,
```
karenc-bq, line 83:
```suggestion
    props: Map<string, any>,
```

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,482 @@
...
+      logger.error(`Error fetching topology: ${error}.`);
+    }
+    return null;
+  }
+
+  private async openAnyClientAndUpdateTopology() {
+    if (!this.monitoringClient) {
+      let client;
+      try {
+        client = await this._pluginService.forceConnect(this.initialHostInfo, this._properties);
```
karenc-bq, line 198: Need to connect using monitoring properties instead of the default properties. Similar comment for other calls to connect

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,482 @@
...
+    this.stopMonitoring = true;
+    this.hostMonitorsStop = true;
+    await Promise.all(this.runPromise ? this.untrackedPromises.concat(this.runPromise) : this.untrackedPromises);
+    this.untrackedPromises = [];
+    this.hostMonitors.clear();
+  }
+
+  async forceMonitoringRefresh(shouldVerifyWriter: boolean, timeoutMs: number): Promise<HostInfo[]> {
+    const currentHosts = this.topologyMap.get(this.clusterId);
+    if (currentHosts) {
```
karenc-bq, line 120: Since currentHosts is either `HostInfo[]` or `null` lets strictly check for null here.

```suggestion
    if (currentHosts !== null) {
```

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,482 @@
...
+  async close(): Promise<void> {
+    this.stopMonitoring = true;
+    this.hostMonitorsStop = true;
+    await Promise.all(this.runPromise ? this.untrackedPromises.concat(this.runPromise) : this.untrackedPromises);
+    this.untrackedPromises = [];
+    this.hostMonitors.clear();
+  }
+
+  async forceMonitoringRefresh(shouldVerifyWriter: boolean, timeoutMs: number): Promise<HostInfo[]> {
+    const currentHosts = this.topologyMap.get(this.clusterId);
```
karenc-bq, line 119: Do we need to check for `ignoreNewTopologyRequestsEndTimeNano` here like in JDBC?

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,482 @@
...
+
+  getWriterHostId(client: ClientWrapper) {
+    return client && client.hostInfo.role === HostRole.WRITER ? client.id : null;
+  }
+
+  async closeConnection(client: ClientWrapper) {
+    if (!client) {
+      return;
+    }
+    await this._pluginService.abortTargetClient(client);
```
karenc-bq, line 250: Why are we using abort here instead of `end`

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,482 @@
...
+    this.requestToUpdateTopology = true;
+
+    if (timeoutMs === 0) {
+      return currentHosts;
+    }
+
+    let timeoutId: any;
+    const timeoutTask: Promise<void> = new Promise((resolve, reject) => {
+      timeoutId = setTimeout(() => {
+        reject("Topology update timed out.");
```
karenc-bq, line 156: Instead of reject lets throw a custom error here similar to failover
```suggestion
        reject(new TopologyUpdateTimeoutError());
```

karenc-bq: Actually, on second thought, instead of writing a new timeoutTask, can we just reuse the `getTimeoutTask()` in utils.ts

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,482 @@
...
+        latestHosts = this.topologyMap.get(this.clusterId);
+        return latestHosts === currentHosts ? null : latestHosts;
+      })
+      .catch((error: any) => {
+        logger.debug(`Could not update topology: ${error}.`);
+        if (JSON.stringify(error).includes("Topology update timed out.")) {
+          throw new AwsWrapperError(`ClusterTopologyMonitor topology update timed out in ${timeoutMs} ms.`);
+        }
+        return null;
+      })
```
karenc-bq, line 171: I think the goal of this method to keep calling `this.topologyMap.get(this.clusterId)` until either current hosts != latest hosts or time out has exceeded. This code seems to return immediately after one call.

Also should not be calling await here: `Promise.race([timeoutTask, await this.topologyUpdated])`.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+  private readonly failoverReaderSuccessCounter: TelemetryCounter;
+  private readonly failoverReaderFailedCounter: TelemetryCounter;
+  private telemetryFailoverAdditionalTopTraceSetting: boolean = false;
+  private _rdsUrlType: RdsUrlType | null = null;
+  private _isInTransaction: boolean = false;
+  private _closedExplicitly: boolean = false;
+  private _lastError: any;
+  /*protected failoverClusterTopologyRefreshRateMsSetting: number = WrapperProperties.FAILOVER_CLUSTER_TOPOLOGY_REFRESH_RATE_MS.defaultValue;
+  protected failoverWriterReconnectIntervalMsSetting: number = WrapperProperties.FAILOVER_WRITER_RECONNECT_INTERVAL_MS.defaultValue;
+  protected failoverReaderConnectTimeoutMsSetting: number = WrapperProperties.FAILOVER_READER_CONNECT_TIMEOUT_MS.defaultValue;*/
```
karenc-bq, line 77: Unused parameters

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+  protected isClosed: boolean = false;
+  failoverMode: FailoverMode | null = null;
+
+  private hostListProviderService?: HostListProviderService;
+  protected enableFailoverSetting: boolean = WrapperProperties.ENABLE_CLUSTER_AWARE_FAILOVER.defaultValue;
+  private failoverTimeoutSettingMs: number = WrapperProperties.FAILOVER_TIMEOUT_MS.defaultValue;
```
karenc-bq, line 83:
```suggestion
  private readonly failoverTimeoutSettingMs: number = WrapperProperties.FAILOVER_TIMEOUT_MS.defaultValue;
```

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+  private telemetryFailoverAdditionalTopTraceSetting: boolean = false;
+  private _rdsUrlType: RdsUrlType | null = null;
+  private _isInTransaction: boolean = false;
+  private _closedExplicitly: boolean = false;
+  private _lastError: any;
+  /*protected failoverClusterTopologyRefreshRateMsSetting: number = WrapperProperties.FAILOVER_CLUSTER_TOPOLOGY_REFRESH_RATE_MS.defaultValue;
+  protected failoverWriterReconnectIntervalMsSetting: number = WrapperProperties.FAILOVER_WRITER_RECONNECT_INTERVAL_MS.defaultValue;
+  protected failoverReaderConnectTimeoutMsSetting: number = WrapperProperties.FAILOVER_READER_CONNECT_TIMEOUT_MS.defaultValue;*/
+  protected isClosed: boolean = false;
+  failoverMode: FailoverMode | null = null;
```
karenc-bq, line 79: Instead of initializing this to null can we use `FailoverMode.UNKNOWN`?

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+    return Promise.resolve(OldConnectionSuggestionAction.NO_OPINION);
+  }
+
+  override async notifyHostListChanged(changes: Map<string, Set<HostChangeOptions>>): Promise<void> {
+    if (!this.enableFailoverSetting) {
+      return Promise.resolve();
+    }
+
+    // Log changes
+    if (logger.level === "debug") {
```
karenc-bq, line 148: Please move this into a private method.
The current implementation will only log the changes if and only if the log level is set to debug. If it is set to something more verbose like `silly` this wont be logged. Please see the implementation in `logutils.ts` to see how to correctly implement this level check.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+      let str = "Changes:";
+      for (const [key, values] of changes.entries()) {
+        str = str.concat("\n");
+        // Convert from int back into enum
+        const valStr = Array.from(values)
+          .map((x) => HostChangeOptions[x])
+          .join(", ");
+        str = str.concat(`\tHost '${key}': ${valStr}`);
```
karenc-bq, line 156:
```suggestion
        str += `\tHost '${key}': ${valStr}`;
```

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+      }
+      logger.debug(str);
```
karenc-bq, line 158: What if there are no changes? Will it just log `Changes:`?

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
karenc-bq, line 1: Maybe I missed it but are there any code to release resources like in `monitorimpl`? Seems like the monitoring promise might run indefinitely and cause the client application to hang.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    logger.debug(Messages.get("Failover.parameterValue", "failoverMode", FailoverMode[this.failoverMode]));
+  }
+
+  override notifyConnectionChanged(changes: Set<HostChangeOptions>): Promise<OldConnectionSuggestionAction> {
+    return Promise.resolve(OldConnectionSuggestionAction.NO_OPINION);
+  }
+
+  override async notifyHostListChanged(changes: Map<string, Set<HostChangeOptions>>): Promise<void> {
```
karenc-bq, line 142: What's the goal of overriding this method? Doesn't seem like we are doing anything other than logigng

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+      // "The active SQL connection has changed. Please re-configure session state if required."
+      logger.debug(Messages.get("Failover.connectionChangedError"));
+      throw new FailoverSuccessError(Messages.get("Failover.connectionChangedError"));
+    }
+    throw new AwsWrapperError(Messages.get("Failover.noOperationsAfterConnectionClosed"));
+  }
+
+  async updateTopology(forceUpdate: boolean) {
+    const client = this.pluginService.getCurrentClient();
+    if (!this.isFailoverEnabled() || !client || !(await client.isValid())) {
```
karenc-bq, line 211: Don't need `!client`, it will never be null or undefined.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+        } catch (error) {
+          if (error instanceof FailoverSuccessError) {
+            client = this.pluginService.getCurrentClient();
+          }
+        }
+      }
+    } else {
+      try {
+        this.pluginService.refreshHostList();
+        this.failover(hostInfo);
```
karenc-bq, line 268:
```suggestion
        await this.pluginService.refreshHostList();
        await this.failover(hostInfo);
```

karenc-bq: Please check all other calls to refreshHostList in this file

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    await this.failover(this.pluginService.getCurrentHostInfo());
+  }
+
+  private isWriter(hostInfo: HostInfo): boolean {
+    return hostInfo.role === HostRole.WRITER;
+  }
+
+  private shouldAttemptReaderConnection(): boolean {
```
karenc-bq, line 537: These two methods are not used

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+        await this.pluginService.abortCurrentClient();
+      }
+    } catch (error) {
+      // swallow this error, current target client should be useless anyway.
+    }
+  }
+
+  async pickNewConnection() {
+    if (this.isClosed && this._closedExplicitly) {
+      logger.debug(Messages.get("Failover.transactionResolutionUnknownError"));
```
karenc-bq, line 526: This log message doesn't seem like the appropriate message, could you please verify with @sergiyvamz

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,575 @@
...
+        if (!(await this.pluginService.initiateTopologyUpdate(false, this.failoverTimeoutSettingMs))) {
+          // Unable to establish SQL connection to an instance.
+          this.failoverReaderFailedCounter.inc();
+          logger.warn(Messages.get("Failover.unableToConnectToReader"));
+          throw new FailoverFailedError(Messages.get("Failover.unableToConnectToReader"));
+        }
+
+        // Signal to connect that this is an internal call and does not require additional processing.
+        const copyProps = new Map<string, any>(this._properties);
+        copyProps.set(Failover2Plugin.INTERNAL_CONNECT_PROPERTY_NAME, true);
```
karenc-bq, line 367: This new property will cause the community MySQL driver to warn 
```
Ignoring invalid configuration option passed to Connection: 76c06979-49c4-4c86-9600-a63605b83f50. 
```
Please make sure it is handled by `removeWrapperProperties`

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
sophia-bq: I renamed my static method to MonitoringRdsHostListProvider.releaseResources() to make it clearer but yes every asynchronous task is tracked and should be closed when this is called at the end of an application.

common/lib/plugins/failover2/failover2_plugin.ts
sophia-bq: Verified and it is does need to be changed, will update on JDBC as well.

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
karenc-bq: Does it make sense to implement it as part of `PluginManager.releaseResources()` so it is more convenient for users?

common/lib/wrapper_property.ts
```diff
@@ -201,6 +202,12 @@ export class WrapperProperties {
     30000
   );
 
+  static readonly CLUSTER_TOPOLOGY_HIGH_REFRESH_RATE_MS = new WrapperProperty<number>(
+    "clusterTopologyRefreshRateMs",
+    "Cluster topology highrefresh rate in millis.",
```
joyc-bq, line 207:
```suggestion
    "Cluster topology refresh rate in millis.",
```

common/lib/connection_plugin_chain_builder.ts
```diff
@@ -83,6 +85,7 @@ export class ConnectionPluginChainBuilder {
     [FederatedAuthPluginFactory, 1200],
     [OktaAuthPluginFactory, 1300],
     [DeveloperConnectionPluginFactory, 1400],
+    [Failover2PluginFactory, 1400],
```
joyc-bq, line 88: should this also be 710?

common/lib/wrapper_property.ts
```diff
@@ -201,6 +202,12 @@ export class WrapperProperties {
     30000
   );
 
+  static readonly CLUSTER_TOPOLOGY_HIGH_REFRESH_RATE_MS = new WrapperProperty<number>(
+    "clusterTopologyRefreshRateMs",
```
joyc-bq, line 206: did you want this to be called `clusterTopologyHighRefreshRateMs`. I see in the docs PR that's what this is called

joyc-bq:
were the integration tests for readwritesplitting + failover also tested with failover2?

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
```diff
@@ -0,0 +1,123 @@
...
+import { AwsWrapperError } from "../../utils/errors";
+import { Messages } from "../../utils/messages";
+import { WrapperProperties } from "../../wrapper_property";
+
+export class MonitoringRdsHostListProvider extends RdsHostListProvider {
+  static readonly CACHE_CLEANUP_NANOS: bigint = BigInt(60_000_000_000); // 1 minute.
+  static readonly MONITOR_EXPIRATION_NANOS: bigint = BigInt(15 * 60_000_000_000); // 15 minutes.
+  static readonly DEFAULT_TOPOLOGY_QUERY_TIMEOUT_MS = 5000; // 5 seconds.
+
+  private static monitors: SlidingExpirationCache<string, ClusterTopologyMonitorImpl> = new SlidingExpirationCache(
```
karenc-bq, line 34:
```suggestion
  private static monitors: SlidingExpirationCache<string, ClusterTopologyMonitor> = new SlidingExpirationCache(
```

https://medium.com/@essaadani.yo/program-to-interface-not-implementation-29154de5c5b4#:~:text=Program%20to%20an%20Interface%2C%20Not,modular%2C%20testable%2C%20and%20scalable.

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,517 @@
...
+import { HostRole } from "../../host_role";
+import { ClientWrapper } from "../../client_wrapper";
+import { AwsWrapperError } from "../../utils/errors";
+import { MonitoringRdsHostListProvider } from "./monitoring_host_list_provider";
+import { Messages } from "../../utils/messages";
+
+export interface ClusterTopologyMonitor {
+  forceRefresh(client: any, timeoutMs: number): Promise<HostInfo[]>;
+
+  close(): void;
```
karenc-bq, line 32:
```suggestion
  close(): Promise<void>;
```

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
```diff
@@ -0,0 +1,123 @@
...
+    super(properties, originalUrl, hostListProviderService);
+    this.pluginService = pluginService;
+  }
+
+  async clearAll(): Promise<void> {
+    RdsHostListProvider.clearAll();
+    for (const [key, monitor] of MonitoringRdsHostListProvider.monitors.entries) {
+      await monitor.item.close();
+    }
+    MonitoringRdsHostListProvider.monitors.clear();
```
karenc-bq, line 58:
```suggestion
    MonitoringRdsHostListProvider.monitors.clear();
```

SlidingExpirationCache's `clear()` method should already be iterating over the entries and calling `close` on them.

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export class MonitoringRdsHostListProvider extends RdsHostListProvider {
+  static readonly CACHE_CLEANUP_NANOS: bigint = BigInt(60_000_000_000); // 1 minute.
+  static readonly MONITOR_EXPIRATION_NANOS: bigint = BigInt(15 * 60_000_000_000); // 15 minutes.
+  static readonly DEFAULT_TOPOLOGY_QUERY_TIMEOUT_MS = 5000; // 5 seconds.
+
+  private static monitors: SlidingExpirationCache<string, ClusterTopologyMonitorImpl> = new SlidingExpirationCache(
+    MonitoringRdsHostListProvider.CACHE_CLEANUP_NANOS,
+    () => true,
+    async (monitor: ClusterTopologyMonitorImpl) => {
```
karenc-bq, line 37:
```suggestion
    async (monitor: ClusterTopologyMonitor) => {
```

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,517 @@
...
+    await Promise.all(this.untrackedPromises);
+    await this.closeConnection(this.monitoringClient);
+    await this.closeConnection(this.hostMonitorsWriterClient);
+    await this.closeConnection(this.hostMonitorsReaderClient);
+    this.untrackedPromises = [];
+    this.hostMonitors.clear();
+  }
+
+  async forceMonitoringRefresh(shouldVerifyWriter: boolean, timeoutMs: number): Promise<HostInfo[]> {
+    if (this.ignoreNewTopologyRequestsEndTime > 0 && Date.now() < this.ignoreNewTopologyRequestsEndTime) {
```
karenc-bq, line 126: Curious if this check `this.ignoreNewTopologyRequestsEndTime > 0` is actually necessary?
I don't think `Date.now()` will ever be <= 0 so I think we can skip this check.

karenc-bq: Same comment for `this.highRefreshRateEndTime`

karenc-bq: Can we add the time unit to both of these variable names? i.e. `highRefreshRateEndTimeMs`

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,517 @@
...
+  protected writerChanged: boolean = false;
+
+  constructor(monitor: ClusterTopologyMonitorImpl, hostInfo: HostInfo, writerHostInfo: HostInfo) {
+    this.monitor = monitor;
+    this.hostInfo = hostInfo;
+    this.writerHostInfo = writerHostInfo;
+  }
+
+  async run() {
+    let client = null;
```
karenc-bq, line 421:
```suggestion
    let client: ClientWrapper | null = null;
```

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,517 @@
...
+      await sleep(1000);
+    }
+
+    if (Date.now() >= endTime) {
+      throw new AwsWrapperError(Messages.get("ClusterTopologyMonitor.timeoutError", timeoutMs.toString()));
+    }
+    return latestHosts;
+  }
+
+  async fetchTopologyAndUpdateCache(client: ClientWrapper): Promise<HostInfo[]> {
```
karenc-bq, line 179:
```suggestion
  async fetchTopologyAndUpdateCache(client: ClientWrapper): Promise<HostInfo[] | null> {
```

karenc-bq: Same comment for other methods that may return null

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,517 @@
...
+import { logTopology, sleep } from "../../utils/utils";
+import { logger } from "../../../logutils";
+import { HostRole } from "../../host_role";
+import { ClientWrapper } from "../../client_wrapper";
+import { AwsWrapperError } from "../../utils/errors";
+import { MonitoringRdsHostListProvider } from "./monitoring_host_list_provider";
+import { Messages } from "../../utils/messages";
+
+export interface ClusterTopologyMonitor {
+  forceRefresh(client: any, timeoutMs: number): Promise<HostInfo[]>;
```
karenc-bq, line 30:
```suggestion
  forceRefresh(client: ClientWrapper, timeoutMs: number): Promise<HostInfo[]>;
```

Avoid using `any` for `client`

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -0,0 +1,517 @@
...
+    await this.closeConnection(this.hostMonitorsReaderClient);
+    this.untrackedPromises = [];
+    this.hostMonitors.clear();
+  }
+
+  async forceMonitoringRefresh(shouldVerifyWriter: boolean, timeoutMs: number): Promise<HostInfo[]> {
+    if (this.ignoreNewTopologyRequestsEndTime > 0 && Date.now() < this.ignoreNewTopologyRequestsEndTime) {
+      // Previous failover has just completed, use results without triggering new update.
+      const currentHosts = this.topologyMap.get(this.clusterId);
+      if (currentHosts !== null) {
```
karenc-bq, line 129: Can we add the `ClusterTopologyMonitorImpl.ignoringTopologyRequest` log here as well?

common/lib/utils/locales/en.json
```diff
@@ -195,5 +198,20 @@
...
   "AwsClient.configurationProfileNotFound": "Configuration profile '%s' not found.",
-  "AwsClient.targetClientNotDefined": "AwsClient targetClient not defined."
+  "AwsClient.targetClientNotDefined": "AwsClient targetClient not defined.",
+  "Failover2.unableToConnect": "Unable to establish SQL connection to the instance.",
+  "Failover2.failoverReaderNotConnectedToReader": "Unable to establish SQL connection to the instance '%s' as a reader.",
+  "Failover2.failoverWriterConnectedToReader": "Unable to establish SQL connection to a writer instance.",
+  "Failover2.unableToFetchTopology": "Unable to establish SQL connection and fetch topology.",
+  "ClusterTopologyMonitor.timeoutError": "ClusterTopologyMonitor topology update timed out in '%s' ms.",
+  "ClusterTopologyMonitor.errorFetchingTopology": "Error fetching topology: '%s'.",
+  "ClusterTopologyMonitor.unableToConnect": "Could not connect to inital host: '%s'.",
```
karenc-bq, line 208:
```suggestion
  "ClusterTopologyMonitor.unableToConnect": "Could not connect to initial host: '%s'.",
```

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
```diff
@@ -0,0 +1,123 @@
...
+  async sqlQueryForTopology(targetClient: ClientWrapper): Promise<HostInfo[]> {
+    const dialect: DatabaseDialect = this.hostListProviderService.getDialect();
+    if (!this.isTopologyAwareDatabaseDialect(dialect)) {
+      throw new TypeError(Messages.get("RdsHostListProvider.incorrectDialect"));
+    }
+    return await dialect.queryForTopology(targetClient, this).then((res: any) => this.processQueryResults(res));
+  }
+
+  async getWriterId(targetClient: ClientWrapper): Promise<string> {
+    const hostInfo: HostInfo = await this.identifyConnection(targetClient, this.hostListProviderService.getDialect());
```
karenc-bq, line 86: `identifyConnection` gets the host name of the current connection, which might not be a writer instance, so I don't think this will reliably return the writer ID.

karenc-bq: Check the `IS_WRITER_QUERY` variable in JDBC

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+import { HostRole } from "../../host_role";
+import { SubscribedMethodHelper } from "../../utils/subscribed_method_helper";
+import { OldConnectionSuggestionAction } from "../../old_connection_suggestion_action";
+import { HostChangeOptions } from "../../host_change_options";
+import { CanReleaseResources } from "../../can_release_resources";
+import { MonitoringRdsHostListProvider } from "../../host_list_provider/monitoring/monitoring_host_list_provider";
+
+export class Failover2Plugin extends AbstractConnectionPlugin implements CanReleaseResources {
+  private static readonly TELEMETRY_WRITER_FAILOVER = "failover to writer instance";
+  private static readonly TELEMETRY_READER_FAILOVER = "failover to replica";
```
karenc-bq, line 49:
```suggestion
  private static readonly TELEMETRY_READER_FAILOVER = "failover to reader";
```

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+  private static readonly METHOD_END = "end";
+  private static readonly subscribedMethods: Set<string> = new Set(["initHostProvider", "connect", "query", "notifyConnectionChanged"]);
```
karenc-bq, line 51:
```suggestion
  private static readonly SUBSCRIBED_METHODS: Set<string> = new Set(["initHostProvider", "connect", "query", "notifyConnectionChanged"]);
```

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+
+    if (this.failoverMode === FailoverMode.UNKNOWN) {
+      this.failoverMode = this._rdsUrlType === RdsUrlType.RDS_READER_CLUSTER ? FailoverMode.READER_OR_WRITER : FailoverMode.STRICT_WRITER;
+    }
+
+    logger.debug(Messages.get("Failover.parameterValue", "failoverMode", FailoverMode[this.failoverMode]));
+  }
+
+  override notifyConnectionChanged(changes: Set<HostChangeOptions>): Promise<OldConnectionSuggestionAction> {
+    return Promise.resolve(OldConnectionSuggestionAction.NO_OPINION);
```
karenc-bq, line 121: This implementation is the same as the one in `AbstractConnectionPlugin`, I don't think we need this here in this file.

common/lib/plugins/failover2/failover2_plugin.ts
karenc-bq: If we are assigning them immediately in the constructor what is the point of setting default values here?
```
  protected enableFailoverSetting: boolean;
  private readonly failoverTimeoutSettingMs: number;
```

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+    initHostProviderFunc: () => void
+  ): void {
+    this.hostListProviderService = hostListProviderService;
+    if (!this.enableFailoverSetting) {
+      return;
+    }
+
+    initHostProviderFunc();
+
+    this.failoverMode = failoverModeFromValue(WrapperProperties.FAILOVER_MODE.get(this._properties));
```
karenc-bq, line 110: Should probably use the `props` passed in instead of `this._properties`

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+    );
+  }
+
+  private async invalidInvocationOnClosedConnection() {
+    if (!this._closedExplicitly) {
+      this.isClosed = false;
+      await this.pickNewConnection();
+
+      // "The active SQL connection has changed. Please re-configure session state if required."
+      logger.debug(Messages.get("Failover.connectionChangedError"));
```
karenc-bq, line 139:
```suggestion
      logger.info(Messages.get("Failover.connectionChangedError"));
```

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+    return (
+      this.enableFailoverSetting &&
+      this._rdsUrlType !== RdsUrlType.RDS_PROXY &&
+      this.pluginService.getHosts() &&
+      this.pluginService.getHosts().length > 0
+    );
+  }
+
+  private async invalidInvocationOnClosedConnection() {
+    if (!this._closedExplicitly) {
```
karenc-bq, line 134: Is `this._closedExplicitly` ever updated? I only see it being initialized to `false`.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+        return await methodFunc();
+      }
+
+      if (this.isClosed) {
+        await this.invalidInvocationOnClosedConnection();
+      }
+
+      if (this.canUpdateTopology(methodName)) {
+        await this.updateTopology();
+      }
```
karenc-bq, line 242: Could you please verify with @sergiyvamz if this is necessary? JDBC's implementation doesn't have this check.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+        this._lastError = e;
+        await this.pickNewConnection();
+      }
+
+      throw e;
+    }
+  }
+
+  async failover(failedHost: HostInfo) {
+    this.pluginService.setAvailability(failedHost.allAliases, HostAvailability.NOT_AVAILABLE); // checking if needed
```
karenc-bq, line 263: Is this needed?

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+      throw new TransactionResolutionUnknownError(Messages.get("Failover.transactionResolutionUnknownError"));
+    } else {
+      // "The active SQL connection has changed due to a connection failure. Please re-configure
+      // session state if required."
+      throw new FailoverSuccessError(Messages.get("Failover.connectionChangedError"));
+    }
+  }
+
+  async failoverReader() {
+    const telemetryFactory = this.pluginService.getTelemetryFactory();
```
karenc-bq, line 284: Looks like Node's implementation is missing some logs, could we double check against JDBC

common/lib/plugin_service.ts
```diff
@@ -176,6 +175,31 @@ export class PluginService implements ErrorHandler, HostListProviderService {
     }
   }
 
+  async initiateTopologyUpdate(shouldVerifyWriter: boolean, timeoutMs: number): Promise<boolean> {
+    const hostListProvider = this.getHostListProvider();
+    if (!(hostListProvider instanceof MonitoringRdsHostListProvider)) {
```
karenc-bq, line 180: Instead of implementing this method that checks for a hard coded class (`hostListProvider instanceof MonitoringRdsHostListProvider`), I recommend implementing a new interface to keep `PluginService`'s implementation generic. JDBC uses this approach as well, see `BlockingHostListProvider`. I also recommend renaming this method to `forceMonitoringRefresh`

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+              readerCandidateClient = null;
+            } catch (err) {
+              readerCandidateClient = null;
+            }
+          }
+        }
+
+        logger.warn(Messages.get("Failover.unableToConnectToReader"));
+        this.failoverReaderFailedCounter.inc();
+        throw new FailoverFailedError(Messages.get("Failover.unableToConnectToReader"));
```
karenc-bq, line 338: Do these need to be in a `catch` block?

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+    const failoverEndTimeMs = Date.now() + this.failoverTimeoutSettingMs;
+
+    try {
+      logger.debug(Messages.get("Failover.startReaderFailover"));
+      await telemetryContext.start(async () => {
+        if (!(await this.pluginService.initiateTopologyUpdate(false, 0))) {
+          // Unable to establish SQL connection to an instance.
+          this.failoverReaderFailedCounter.inc();
+          logger.warn(Messages.get("Failover2.unableToFetchTopology"));
+          throw new FailoverFailedError(Messages.get("Failover.unableToConnectToReader"));
```
karenc-bq, line 298: Should the messages for these match?
Also let's use `logger.error` here since the workflow failed.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+        const copyProps = new Map<string, any>(this._properties);
+        copyProps.set(Failover2Plugin.INTERNAL_CONNECT_PROPERTY_NAME, true);
+
+        while (Date.now() < failoverEndTimeMs) {
+          const hosts: HostInfo[] = this.pluginService.getHosts();
+          const hostsByPriority = this.getHostsByPriority(hosts);
+          let readerCandidateClient: ClientWrapper = null;
+          let readerCandidateHostInfo: HostInfo = null;
+
+          while (readerCandidateClient === null && hostsByPriority.length > 0 && Date.now() < failoverEndTimeMs) {
```
karenc-bq, line 311:
```suggestion
          while (hostsByPriority.length > 0 && Date.now() < failoverEndTimeMs) {
```
`readerCandidateClient` will always be `null` when we get to this condition.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+          throw new FailoverFailedError(Messages.get("Failover.unableToConnectToReader"));
+        }
+
+        // Signal to connect that this is an internal call and does not require additional processing.
+        const copyProps = new Map<string, any>(this._properties);
+        copyProps.set(Failover2Plugin.INTERNAL_CONNECT_PROPERTY_NAME, true);
+
+        while (Date.now() < failoverEndTimeMs) {
+          const hosts: HostInfo[] = this.pluginService.getHosts();
+          const hostsByPriority = this.getHostsByPriority(hosts);
```
karenc-bq, line 307: Curious why are we using `getHostsByPriority` here?

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+          let readerCandidateClient: ClientWrapper = null;
+          let readerCandidateHostInfo: HostInfo = null;
+
+          while (readerCandidateClient === null && hostsByPriority.length > 0 && Date.now() < failoverEndTimeMs) {
+            readerCandidateHostInfo = hostsByPriority.shift();
```
karenc-bq, line 312: Could you please verify the implementation for `readerCandidateHostInfo`? Is it supposed be from `getHostsByPriority` or `getHostInfoByStrategy`?
Also this class seems to be missing the connection parameter: `FAILOVER_READER_HOST_SELECTOR_STRATEGY`

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+              readerCandidateClient = await this.pluginService.connect(readerCandidateHostInfo, copyProps);
+              if (
+                readerCandidateClient &&
+                ((await this.pluginService.getHostRole(readerCandidateClient)) === HostRole.READER || this.failoverMode != FailoverMode.STRICT_READER)
+              ) {
+                logger.info(Messages.get("Failover.establishedConnection", readerCandidateHostInfo.host));
+                this.pluginService.getCurrentHostInfo()?.removeAlias(Array.from(oldAliases));
+                await this.pluginService.abortCurrentClient();
+                await this.pluginService.setCurrentClient(readerCandidateClient, readerCandidateHostInfo);
+                await this.updateTopology();
```
karenc-bq, line 323: Why do we need to call `updateTopology` here? Same comment goes for the other call in `failoverWriter`.
Shouldn't the earlier call `this.pluginService.initiateTopologyUpdate` be sufficient?

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+    const telemetryContext = telemetryFactory.openTelemetryContext(Failover2Plugin.TELEMETRY_WRITER_FAILOVER, TelemetryTraceLevel.NESTED);
+    this.failoverWriterTriggeredCounter.inc();
+
+    try {
+      logger.debug(Messages.get("Failover.startWriterFailover"));
+      await telemetryContext.start(async () => {
+        if (!(await this.pluginService.initiateTopologyUpdate(true, this.failoverTimeoutSettingMs))) {
+          // Unable to establish SQL connection to writer node.
+          this.failoverWriterFailedCounter.inc();
+          logger.warn(Messages.get("Failover2.unableToFetchTopology"));
```
karenc-bq, line 358:
```suggestion
          logger.error(Messages.get("Failover2.unableToFetchTopology"));
```

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,502 @@
...
+        if (writerCandidateHostInfo) {
+          try {
+            writerCandidateClient = await this.pluginService.connect(writerCandidateHostInfo, copyProps);
+          } catch (err) {
+            // Do nothing.
+          }
+        }
+
+        if (!writerCandidateClient) {
+          logger.warn(Messages.get("Failover.unableToConnectToWriter"));
```
karenc-bq, line 381:
```suggestion
          logger.error(Messages.get("Failover.unableToConnectToWriter"));
```

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,497 @@
...
+  }
+
+  async failover() {
+    if (this.failoverMode === FailoverMode.STRICT_WRITER) {
+      await this.failoverWriter();
+    } else {
+      await this.failoverReader();
+    }
+
+    if (this._isInTransaction || this.pluginService.isInTransaction()) {
```
karenc-bq, line 245: `this._isInTransaction` is set via `this.pluginService.isInTransaction()`, and when we use it we check `this.pluginService.isInTransaction()` anyway. Not sure if it is worth having this local variable.

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -0,0 +1,497 @@
...
+            logger.error(Messages.get("Failover.unableToConnectToWriter"));
+            this.failoverWriterFailedCounter.inc();
+            throw new FailoverFailedError(Messages.get("Failover.unableToConnectToWriter"));
+          }
+        }
+
+        if (!writerCandidateClient) {
+          logger.error(Messages.get("Failover.unableToConnectToWriter"));
+          this.failoverWriterFailedCounter.inc();
+          throw new FailoverFailedError(Messages.get("Failover.unableToConnectToWriter"));
```
karenc-bq, line 414: Can we refactor this logic into a method to reduce duplicated code
```
  private logAndThrowError(errorMessage: string) {
    logger.error(errorMessage);
    this.failoverWriterFailedCounter.inc();
    throw new FailoverFailedError(errorMessage);
  }
```

karenc-bq (approved):

common/lib/plugins/failover2/failover2_plugin.ts
sophia-bq: this._isInTransaction is set right before we rollback the client. It stores the knowledge that we are in transaction, as rollback can set inTransaction to be false

common/lib/plugin_service.ts
```diff
@@ -176,6 +174,34 @@ export class PluginService implements ErrorHandler, HostListProviderService {
...
+      }
+    } catch (err) {
+      // Do nothing.
+      logger.info(Messages.get("PluginService.forceMonitoringRefreshTimeout", timeoutMs.toString()));
+    }
+
+    return false;
+  }
+
+  protected isBlockingHostListProvider(arg: any): arg is BlockingHostListProvider {
```
crystall-bitquill, line 201: Would it make more sense to put this in utils?

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
```diff
@@ -0,0 +1,114 @@
...
+  async queryForTopology(targetClient: ClientWrapper, dialect: DatabaseDialect): Promise<HostInfo[]> {
+    const monitor: ClusterTopologyMonitor = this.initMonitor();
+
+    try {
+      return await monitor.forceRefresh(targetClient, MonitoringRdsHostListProvider.DEFAULT_TOPOLOGY_QUERY_TIMEOUT_MS);
+    } catch (error) {
+      logger.info(Messages.get("MonitoringHostListProvider.errorForceRefresh", error.message));
+      return null;
+    }
+
```
crystall-bitquill, line 75:
```suggestion
```

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
```diff
@@ -0,0 +1,114 @@
...
+  async forceMonitoringRefresh(shouldVerifyWriter: boolean, timeoutMs: number): Promise<HostInfo[]> {
+    const monitor: ClusterTopologyMonitor = this.initMonitor();
+
+    return await monitor.forceMonitoringRefresh(shouldVerifyWriter, timeoutMs);
+  }
+
+  protected initMonitor(): ClusterTopologyMonitor {
+    const monitor: ClusterTopologyMonitor = MonitoringRdsHostListProvider.monitors.computeIfAbsent(
+      this.clusterId,
+      (x) =>
```
crystall-bitquill, line 95:
```suggestion
      () =>
```

crystall-bitquill (approved):

joyc-bq (approved):

common/lib/plugin_service.ts
sophia-bq: Hmm I don't think it belongs in utils, it is very specific to this one use

PR #366, opened 2025-01-09
More robust MySQL catalog usage detection

adrianfalleiro:
### Summary

The current MySQL catalog detection uses substring match to detect if a query makes use of a `USE` statement. The usage of substring match is brittle and fails if any part of the statement (other than a comment) has the substring `use`. 

This is problematic as `user` is a very common table name and gets mistakenly detected.

```ts
// Should return `undefined` but returns `"*"`
SqlMethodUtils.doesSetCatalog("select * from user", new MySQLDatabaseDialect()) // => "*"
```

### Description

This PR changes the substring match to a more robust regular expression. I also added a few tests cases.


sergiyvamz (approved):

crystall-bitquill (approved):

crystall-bitquill:
Thanks for your PR!

PR #375, opened 2025-01-20
Passthrough source error as `cause` to `AwsWrapperError` constructor

adrianfalleiro:
### Summary

Passes through the source error from the pg/mysql driver to the `AwsWrapperError` constructor. 

### Description

I am currently using the `mysql2` library and looking to integrate this wrapper with my project. The exceptions thrown by `mysql2` include extra details that are useful for handling the error (code, errno, sqlState, sqlMessage and sql). 

When this library throws an error, these details are lost because the error is wrapped in an `AwsWrapperError` and only the message is preserved. I noticed that this class accepts an optional `cause` in the constructor...but it's not being passed in `ClientUtils.queryWithTimeout()`. This PR fixes that.

Didn't add a test as I was unsure where it would belong. Happy to add one.


adrianfalleiro:
@karenc-bq @crystall-bitquill @sergiyvamz Tagging y'all because you reviewed and merged my previous PR. Thanks for looking at my PR.

sophia-bq (approved):
Hi @adrianfalleiro, 

Thank you for reaching out and raising this. Looks like a good change to me.

sophia-bq:
Hi @adrianfalleiro, 

Thank you once again for your PR. I'd be happy to merge this change in for you, the branch just needs to be up to date with main. Would you like to update the branch or shall I?

adrianfalleiro:
@sophia-bq thanks! Just updated it

PR #390, opened 2025-01-30
fix: reader failover wait for complete batch

crystall-bitquill:
### Summary

Fix connections being unclosed in reader failover handler

### Description

- Currently connection attempts are made in batches of two, and the first settled promise is returned. However, if the first connection task fails, the second is also abandoned. This PR changes the reader failover implementation to wait for the entire batch to complete before moving on.
- This PR also changes the failover plugin to initialize the failover settings prior to creating the writer and reader failover handlers.

### DIFF
--- a/common/lib/plugins/failover/reader_failover_handler.ts
+++ b/common/lib/plugins/failover/reader_failover_handler.ts
@@ -152,13 +152,9 @@ export class ClusterAwareReaderFailoverHandler implements ReaderFailoverHandler 
           return result;
         }
       } catch (error) {
-        if (error instanceof AggregateError && error.message.includes("All promises were rejected")) {
-          // ignore and try the next batch
-        } else {
-          // Failover has failed.
-          this.taskHandler.setSelectedConnectionAttemptTask(failoverTaskId, ClusterAwareReaderFailoverHandler.FAILOVER_FAILED);
-          throw error;
-        }
+        // Failover has failed.
+        this.taskHandler.setSelectedConnectionAttemptTask(failoverTaskId, ClusterAwareReaderFailoverHandler.FAILOVER_FAILED);
+        throw error;
       }
 
       await sleep(1000);
@@ -184,7 +180,7 @@ export class ClusterAwareReaderFailoverHandler implements ReaderFailoverHandler 
         throw new AwsWrapperError("Connection attempt task timed out.");
       })
       .catch((error) => {
-        if (error instanceof InternalQueryTimeoutError || (error instanceof AggregateError && error.message.includes("All promises were rejected"))) {
+        if (error instanceof InternalQueryTimeoutError) {
           // ignore so the next task batch can be attempted
           return ClusterAwareReaderFailoverHandler.FAILED_READER_FAILOVER_RESULT;
         }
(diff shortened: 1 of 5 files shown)


common/lib/plugins/failover/reader_failover_handler.ts
```diff
@@ -207,7 +207,27 @@ export class ClusterAwareReaderFailoverHandler implements ReaderFailoverHandler
...
+            "ClusterAwareReaderFailoverHandler.batchFailed",
+            `[${hosts[i].hostId}${numTasks === 2 ? `, ${hosts[i + 1].hostId}` : ``}]`,
+            `[\n${errors}\n]`
+          )
+        );
+        return new ReaderFailoverResult(null, null, false, awsWrapperError);
+      } else {
+        return new ReaderFailoverResult(null, null, false, undefined);
+      }
+    });
```
karenc-bq, line 230:
```suggestion
      }
      return new ReaderFailoverResult(null, null, false, undefined);
    });
```

common/lib/plugins/failover/reader_failover_handler.ts
```diff
@@ -207,7 +207,27 @@ export class ClusterAwareReaderFailoverHandler implements ReaderFailoverHandler
       tasks.push(secondTask.call());
     }
 
-    return await Promise.any(tasks);
+    return await Promise.any(tasks).catch((error: AggregateError) => {
+      let errors: string = "";
+      for (const e of error.errors) {
+        // Propagate errors that are not caused by network errors.
+        if (!this.pluginService.isNetworkError(error)) {
```
karenc-bq, line 214: What is the goal of this check here? Are we just checking to see if the aggregate error contains the AwsWrapperError we throw to indicate task aborted?
```
throw new AwsWrapperError("Selected task has been chosen. Abort client for host: " + this.newHost.host);
```

In this case would it be more clear if we just check for `error instanceof AwsWrapperError`?

common/lib/plugins/failover/reader_failover_handler.ts
```diff
@@ -207,7 +207,27 @@ export class ClusterAwareReaderFailoverHandler implements ReaderFailoverHandler
...
+          errors += `\n\t${e} - ${e.message}`;
+        }
+      }
+      if (errors) {
+        const awsWrapperError = new AwsWrapperError(
+          Messages.get(
+            "ClusterAwareReaderFailoverHandler.batchFailed",
+            `[${hosts[i].hostId}${numTasks === 2 ? `, ${hosts[i + 1].hostId}` : ``}]`,
+            `[\n${errors}\n]`
```
karenc-bq, line 223: Curious if we need to build this custom message or if we could just use the aggregate error's message

common/lib/plugins/failover/reader_failover_handler.ts
```diff
@@ -342,6 +353,7 @@ class ConnectionAttemptTask {
   async performFinalCleanup() {
     if (this.taskHandler.getSelectedConnectionAttemptTask(this.failoverTaskId) !== this.taskId) {
       await this.pluginService.abortTargetClient(this.targetClient);
+      this.taskHandler.setSelectedConnectionAttemptTask(this.failoverTaskId, -1);
```
karenc-bq, line 356: Why do we need to reset the failoverTaskId here?

common/lib/plugins/failover/reader_failover_handler.ts
karenc-bq: We could also be more specific and create a new `ReaderTaskAbortedError`

common/lib/plugins/failover/reader_failover_handler.ts
```diff
@@ -321,18 +341,9 @@ class ConnectionAttemptTask {
         this.taskHandler.setSelectedConnectionAttemptTask(this.failoverTaskId, this.taskId);
         return new ReaderFailoverResult(this.targetClient, this.newHost, true, undefined, this.taskId);
       }
-      await this.pluginService.abortTargetClient(this.targetClient);
-      return new ReaderFailoverResult(null, null, false, undefined, this.taskId);
+      throw new AwsWrapperError("Selected task has been chosen. Abort client for host: " + this.newHost.host);
```
karenc-bq, line 344: Should add this message to `en.json`

common/lib/plugins/failover/reader_failover_handler.ts
crystall-bitquill: If we return a ReaderFailoverResult with an error here it will be sent back to the user, so I'd prefer for them to see an AwsWrapperError with some additional context rather than an AggregateError

common/lib/plugins/failover/reader_failover_handler.ts
crystall-bitquill: In this check we want to ignore any network errors so we can continue trying connections to other instances (for example, `connect ECONNREFUSED`). If there are non-network errors, they will be propagated back to the user through the ReaderFailoverResult. This catch block should only be hit when both connection attempts in the batch failed, but the selected task error is only thrown when there has been a successful connection.

karenc-bq (approved):

PR #393, opened 2025-02-03
fix: setQueryTimeout for mysql2DriverDialect

joyc-bq:

### DIFF
--- a/mysql/lib/dialect/mysql2_driver_dialect.ts
+++ b/mysql/lib/dialect/mysql2_driver_dialect.ts
@@ -71,7 +71,7 @@ export class MySQL2DriverDialect implements DriverDialect {
   setQueryTimeout(props: Map<string, any>, sql?: any, wrapperConnectTimeout?: any) {
     const timeout = wrapperConnectTimeout ?? props.get(WrapperProperties.WRAPPER_QUERY_TIMEOUT.name);
     if (timeout && !sql[MySQL2DriverDialect.QUERY_TIMEOUT_PROPERTY_NAME]) {
-      sql[MySQL2DriverDialect.QUERY_TIMEOUT_PROPERTY_NAME] = timeout;
+      sql[MySQL2DriverDialect.QUERY_TIMEOUT_PROPERTY_NAME] = Number(timeout);
     }
   }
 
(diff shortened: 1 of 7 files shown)


mysql/lib/dialect/mysql2_driver_dialect.ts
```diff
@@ -71,7 +71,7 @@ export class MySQL2DriverDialect implements DriverDialect {
   setQueryTimeout(props: Map<string, any>, sql?: any, wrapperConnectTimeout?: any) {
```
crystall-bitquill, line 71:
```suggestion
  setQueryTimeout(props: Map<string, any>, sql?: any, wrapperQueryTimeout?: any) {
```

mysql/lib/dialect/mysql_database_dialect.ts
```diff
@@ -118,17 +118,14 @@ export class MySQLDatabaseDialect implements DatabaseDialect {
...
-          .query({ sql: "SELECT 1" })
-          .then(() => {
-            return true;
-          })
-          .catch(() => {
-            return false;
-          }),
-        targetClient.properties
-      );
+      return await targetClient
```
crystall-bitquill, line 121: Is queryWithTimeout not necessary here?

pg/lib/dialect/node_postgres_driver_dialect.ts
```diff
@@ -71,7 +71,7 @@ export class NodePostgresDriverDialect implements DriverDialect {
     }
   }
 
-  setQueryTimeout(props: Map<string, any>, sql?: any, wrapperQueryTimeout?: any) {
+  setQueryTimeout(props: Map<string, any>, wrapperQueryTimeout?: any) {
```
karenc-bq, line 74: Need to revert this change since we are going with the new fix that does not require changing the interface

karenc-bq (approved):

mysql/lib/dialect/mysql_database_dialect.ts
joyc-bq: we set timeout parameters in query() already, but will revert the changes and keep setQueryTimeout because it has other things going on like deprecated properties checks etc

PR #395, opened 2025-02-03
fix: error retrieving keep alive settings

karenc-bq:
### Summary

Calling `keepAliveProps.get(...)` will throw errors instead of returning undefined because `keepAliveProps` is not a map object.

### DIFF
--- a/mysql/lib/dialect/mysql2_driver_dialect.ts
+++ b/mysql/lib/dialect/mysql2_driver_dialect.ts
@@ -76,6 +76,10 @@ export class MySQL2DriverDialect implements DriverDialect {
   }
 
   setKeepAliveProperties(props: Map<string, any>, keepAliveProps: any) {
+    if (keepAliveProps instanceof Map) {
+      keepAliveProps = Object.fromEntries(keepAliveProps);
+    }
+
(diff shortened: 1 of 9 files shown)


sophia-bq (approved):

mysql/lib/dialect/mysql2_driver_dialect.ts
```diff
@@ -76,7 +76,7 @@ export class MySQL2DriverDialect implements DriverDialect {
   }
 
   setKeepAliveProperties(props: Map<string, any>, keepAliveProps: any) {
```
sophia-bq, line 78: Is keepAliveProps a consistent type? Could add that instead of having `any`?

karenc-bq: I decided to keep it as any so users can pass the keepAliveProps as either a map or an object:
```ts
  const keepAliveProps = new Map<string, any>([
      ["keepAlive", true],
      ["keepAliveInitialDelayMillis", 1234]
    ]);

  const keepAliveObj = {
    keepAlive: true,
    keepAliveInitialDelayMillis: 1234
  };
```

PR #416, opened 2025-03-11
fix: check if first connection after failover is stale

karenc-bq:
### Summary

After failover, if the driver reconnects to the initial connection and use it to query the current topology, it may return a stale topology indicating the current connection is still a writer when it has been switched to a reader.

We need to add extra checks to verify if we have the correct writer.


common/lib/utils/locales/en.json
```diff
@@ -211,7 +212,7 @@
   "AwsClient.targetClientNotDefined": "AwsClient targetClient not defined.",
   "Failover2.unableToConnect": "Unable to establish SQL connection to the instance.",
```
sophia-bq, line 213:
```suggestion
```
The only use of this message is removed in this PR

sophia-bq (approved):

joyc-bq (approved):

PR #426, opened 2025-03-27
test: secrets manager integration

joyc-bq:

### DIFF
--- a/tests/integration/container/tests/secrets_manager.test.ts
+++ b/tests/integration/container/tests/secrets_manager.test.ts
@@ -44,7 +44,7 @@ let secretARN: string;
 let secretsManagerClient: RDSClient;
 let auroraTestUtility: AuroraTestUtility;
 
-async function initDefaultConfig(host: string): Promise<any> {
+async function initConfig(host: string, secretId: string): Promise<any> {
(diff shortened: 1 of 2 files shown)


tests/integration/container/tests/secrets_manager.test.ts
```diff
@@ -0,0 +1,236 @@
...
+let auroraTestUtility: AuroraTestUtility;
+
+async function initDefaultConfig(host: string): Promise<any> {
+  env = await TestEnvironment.getCurrent();
+  let props = {
+    host: host,
+    database: env.databaseInfo.defaultDbName,
+    port: env.databaseInfo.instanceEndpointPort,
+    secretRegion: env.region,
+    secretId: secretId,
```
sophia-bq, line 54: The only difference between initDefaultConfig and initSecretARNConfig is the value passed to secretId. Additionally in several of the tests the value of secretId in the properties is overwritten to something that is not the secretId variable set by createCommand. One potential option is to refactor so initConfig takes in a secretId variable and then you can set it once for every test and don't need initSecretARNConfig

sophia-bq (approved):

PR #438, opened 2025-04-07
test: initial connection strategy integration test

joyc-bq:

### DIFF
--- a/tests/integration/container/tests/initial_connection_strategy.test.ts
+++ b/tests/integration/container/tests/initial_connection_strategy.test.ts
@@ -40,6 +40,7 @@ let driver;
 let initClientFunc: (props: any) => any;
 let client: any;
 let auroraTestUtility: AuroraTestUtility;
+let numReaders;
 
 async function initConfig(readerHostSelectorStrategy: string): Promise<any> {
   let config: any = {
@@ -75,6 +76,7 @@ describe("aurora initial connection strategy", () => {
 
     RdsHostListProvider.clearAll();
     PluginService.clearHostAvailabilityCache();
+    numReaders = env.databaseInfo.instances.length - 1;
   }, 1320000);
 
   afterEach(async () => {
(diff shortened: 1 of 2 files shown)


tests/integration/container/tests/initial_connection_strategy.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    "test round robin host weight pairs",
+    async () => {
+      const config = await initConfig("roundRobin");
+      const numReaders = env.databaseInfo.instances.length - 1;
+      const connectedReaderIds: Set<string> = new Set();
+      const connectionsSet: Set<any> = new Set();
+      try {
+        for (let i = 0; i < numReaders; i++) {
+          client = initClientFunc(config);
+          config["roundRobinHostWeightPairs"] = "instance-2:5";
```
sophia-bq, line 128: I believe that adjusting config after you've initiated the client has no impact so really line 128 impacts the first client from 137 and then you assign the same value again in config for no new impact on line 139. I think it might be cleaner to place this line right before you use config for a client you want this to apply to - in the case of this test just once right before the for loop. Not sure if this requires more adjusting than that to preserve the intent of your test?

tests/integration/container/tests/initial_connection_strategy.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+    await PluginManager.releaseResources();
+    logger.info(`Test finished: ${expect.getState().currentTestName}`);
+  }, 1320000);
+
+  itIf(
+    "test round robin",
+    async () => {
+      const config = await initConfig("roundRobin");
```
sophia-bq, line 95: It might also be worthwhile to have one test that uses a strategy that is not roundRobin just to show it is possible :)

tests/integration/container/tests/initial_connection_strategy.test.ts
```diff
@@ -0,0 +1,154 @@
...
+          await client.connect();
+
+          const readerId = await auroraTestUtility.queryInstanceId(client);
+          expect(connectedReaderIds).not.toContain(readerId);
+          connectedReaderIds.add(readerId);
+          connectionsSet.add(client);
+        }
+        for (let i = 0; i < 3; i++) {
+          client = initClientFunc(config);
+          // all remaining connections should be made to instance with highest weight pair
```
sophia-bq, line 138:
```suggestion
          // All remaining connections should be made to instance with highest weight pair.
```

tests/integration/container/tests/initial_connection_strategy.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  itIfMinThreeInstance(
+    "test round robin host weight pairs",
+    async () => {
+      const connectedReaderIds: Set<string> = new Set();
+      const connectionsSet: Set<any> = new Set();
+      const initialReader = env.databaseInfo.readerInstanceId;
+      const readerCount = instanceCount - 1;
+      const config = await initConfig("roundRobin");
+      config["roundRobinHostWeightPairs"] = `${initialReader}:${readerCount}`;
+      logger.debug("initial reader: " + initialReader);
```
sophia-bq, line 133: Is this intended to be kept or was it for debugging?

sophia-bq (approved):

PR #446, opened 2025-04-29
fix: round robin strategy initial host list is cluster endpoint

joyc-bq:
### Description

In the InitialConnectionStrategyPlugin, in getVerifiedReaderClient when getReader is called, this causes us to get our host with the roundRobin strategy via the getHost method.

 On the first connection, we are passing in the cluster endpoint into hosts ` getHost(hosts: HostInfo[], role: HostRole, props?: Map<string, any>)`. It is not until we call ` const readerCandidate = await this.pluginService.identifyConnection(readerCandidateClient);` that we actually know the list of hosts. 

So when roundRobin Host Selector looks for the list for eligibleHosts upon the initial connection, it receives the cluster endpoint.

By moving the call to getReader after the readerCandidate is retrieved, `getHost(hosts: HostInfo[], role: HostRole, props?: Map<string, any>) ` now passes in the host list which contains the list of eligible instances

### DIFF
--- a/common/lib/plugins/aurora_initial_connection_strategy_plugin.ts
+++ b/common/lib/plugins/aurora_initial_connection_strategy_plugin.ts
@@ -191,16 +191,13 @@ export class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
       readerCandidate = null;
 
       try {
+        readerCandidateClient = await connectFunc();
+        await this.pluginService.forceRefreshHostList(readerCandidateClient);
+        readerCandidate = await this.pluginService.identifyConnection(readerCandidateClient);
         readerCandidate = this.getReader(props);
         // convert into null if undefined
         readerCandidate = readerCandidate ?? null;
-
         if (readerCandidate === null || this.rdsUtils.isRdsClusterDns(readerCandidate.host)) {
-          // Reader is not found. It seems that topology is outdated.
-          readerCandidateClient = await connectFunc();
-          await this.pluginService.forceRefreshHostList(readerCandidateClient);
-          readerCandidate = await this.pluginService.identifyConnection(readerCandidateClient);
-
           if (readerCandidate) {
             if (readerCandidate.role !== HostRole.READER) {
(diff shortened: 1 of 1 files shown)


common/lib/plugins/aurora_initial_connection_strategy_plugin.ts
```diff
@@ -191,16 +191,13 @@ export class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
       readerCandidate = null;
 
       try {
+        readerCandidateClient = await connectFunc();
```
sergiyvamz, line 194: In some case this client can be disposed without proper closing it. For example, at line 224, a new client is created and that overrides existing client.

common/lib/plugins/aurora_initial_connection_strategy_plugin.ts
```diff
@@ -191,16 +191,13 @@ export class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
...
+        await this.pluginService.forceRefreshHostList(readerCandidateClient);
```
sergiyvamz, line 195: Topology is refreshed on every iteration that isn't good.

common/lib/plugins/aurora_initial_connection_strategy_plugin.ts
```diff
@@ -200,7 +200,9 @@ export class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
           readerCandidateClient = await connectFunc();
           await this.pluginService.forceRefreshHostList(readerCandidateClient);
           readerCandidate = await this.pluginService.identifyConnection(readerCandidateClient);
-
+          if (isInitialConnection) {
```
sergiyvamz, line 203: So we're overriding results of the previous line in case it's an initial connection? I'm not sure I follow the logic.

joyc-bq:
No fix as this introduces potential performance concerns. Will update documentation and integration tests in a new PR

PR #472, opened 2025-05-30
fix: postgres instances not up for read write splitting on latest

joyc-bq:
### Description
fixed by adding ssl certificate to the parameters when connecting to the postgres instance


tests/integration/container/tests/read_write_splitting.test.ts
```diff
@@ -31,6 +31,7 @@ import { HostInfo } from "../../../../common/lib/host_info";
 import { PluginManager } from "../../../../common/lib";
 import { RdsHostListProvider } from "../../../../common/lib/host_list_provider/rds_host_list_provider";
 import { PluginService } from "../../../../common/lib/plugin_service";
+import { readFileSync } from "fs";
```
crystall-bitquill, line 34: Is this import needed?

PR #498, opened 2025-07-04
feat: blue green support

karenc-bq:


crystall-bitquill:
Are there unit tests for the new bg classes?

tests/integration/host/src/test/java/integration/host/TestEnvironmentConfig.java
```diff
@@ -197,7 +370,35 @@ private static void createDatabaseContainers(TestEnvironmentConfig env) {
...
+    env.rdsDbRegion = System.getenv("RDS_DB_REGION");
+    env.info.setRegion(
+        !StringUtils.isNullOrEmpty(env.rdsDbRegion)
+            ? env.rdsDbRegion
+            : "us-east-2");
+
+    env.reuseDb = Boolean.parseBoolean(System.getenv("REUSE_RDS_DB"));
+    env.rdsDbName = System.getenv("RDS_DB_NAME"); // "cluster-mysql", "instance-name", "cluster-multi-az-name"
+    env.rdsDbDomain = System.getenv("RDS_DB_DOMAIN"); // "XYZ.us-west-2.rds.amazonaws.com"
+    env.rdsEndpoint = System.getenv("RDS_ENDPOINT"); // "XYZ.us-west-2.rds.amazonaws.com"
```
aaron-congo, line 383:
```suggestion
    env.rdsEndpoint = System.getenv("RDS_ENDPOINT"); // "https://rds-int.amazon.com"
```

common/lib/plugins/bluegreen/blue_green_plugin.ts
```diff
@@ -0,0 +1,224 @@
...
+export interface BlueGreenProviderSupplier {
+  create(pluginService: PluginService, props: Map<string, any>, bgdId: string): BlueGreenStatusProvider;
+}
+
+export class BlueGreenPlugin extends AbstractConnectionPlugin {
+  private static readonly SUBSCRIBED_METHODS: Set<string> = new Set([
+    // We should NOT subscribe to "forceConnect" pipeline since it's used by
+    // BG monitoring, and we don't want to intercept/block those monitoring connections.
+    "connect",
+    "query"
```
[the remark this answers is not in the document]
karenc-bq: Yes, there isn't a universal database API in Node so there are methods like prepareStatement.

common/lib/plugins/bluegreen/blue_green_status_provider.ts
```diff
@@ -0,0 +1,897 @@
...
+                this.rdsUtils.isRdsCustomClusterDns(host) &&
+                customClusterName === this.rdsUtils.removeGreenInstancePrefix(this.rdsUtils.getRdsClusterId(host))
+              );
+            });
+            if (custom.length > 0) {
+              const host = custom[0];
+              if (!this.correspondingHosts.has(blueHost)) {
+                this.correspondingHosts.set(
+                  blueHost,
+                  new Pair(this.hostInfoBuilder.withHost(host).build(), this.hostInfoBuilder.withHost(host).build())
```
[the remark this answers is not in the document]
karenc-bq: Good catch, should fix this on JDBC as well

common/lib/plugins/bluegreen/blue_green_plugin.ts
```diff
@@ -0,0 +1,236 @@
...
+      if (!routing) {
+        return this.regularOpenConnection(connectFunc, isInitialConnection);
+      }
+
+      this.startTimeNano = getTimeInNanos();
+      while (routing && !client) {
+        client = await routing.apply(this, hostInfo, props, isInitialConnection, connectFunc, this.pluginService);
+        if (!client) {
+          this.bgStatus = this.pluginService.getStatus<BlueGreenStatus>(BlueGreenStatus, this.bgdId);
+          if (! this.bgStatus) {
```
aaron-congo, line 116:
```suggestion
          if (!this.bgStatus) {
```

common/lib/plugins/bluegreen/blue_green_plugin.ts
```diff
@@ -0,0 +1,236 @@
...
+        routing.isMatch(currentHostInfo, hostRole)
+      )[0];
+
+      if (!routing) {
+        return await methodFunc();
+      }
+
+      this.startTimeNano = getTimeInNanos();
+
+      while (routing && !result) {
```
aaron-congo, line 178: In JDBC `result` is an Optional so that we can distinguish between when the execute routing did not call the jdbc method (in which case the result will be Optional.EMPTY) versus when the method was called but it was void or returned null (in which case the result will be Optional.of(null)). Do we need to do the same here? We only want to enter this block if the method was not called, and !result evaluates to true if result is null or undefined

common/lib/plugins/bluegreen/blue_green_plugin.ts
```diff
@@ -0,0 +1,236 @@
...
+      )[0];
+
+      if (!routing) {
+        return this.regularOpenConnection(connectFunc, isInitialConnection);
+      }
+
+      this.startTimeNano = getTimeInNanos();
+      while (routing && !client) {
+        client = await routing.apply(this, hostInfo, props, isInitialConnection, connectFunc, this.pluginService);
+        if (!client) {
```
aaron-congo, line 114: nit: you can invert this to get rid of a level of nesting
```suggestion
        if (client) {
          break
        }
```

common/lib/plugins/bluegreen/blue_green_plugin.ts
```diff
@@ -0,0 +1,236 @@
...
+        return await methodFunc();
+      }
+
+      this.startTimeNano = getTimeInNanos();
+
+      while (routing && !result) {
+        result = await routing.apply(this, methodName, methodFunc, methodArgs, this.properties, this.pluginService);
+        if (!result) {
+          this.bgStatus = this.pluginService.getStatus<BlueGreenStatus>(BlueGreenStatus, this.bgdId);
+          routing = this.bgStatus.executeRouting.filter((routing: ExecuteRouting) => routing.isMatch(currentHostInfo, hostRole))[0];
```
aaron-congo, line 182: The bug I asked you to port over from JDBC for connect() is similar to another bug fix in that same PR that needs to be ported over as well, if the status is null we should set the end time and then call the methodFunc

common/lib/plugins/bluegreen/blue_green_status_provider.ts
```diff
@@ -0,0 +1,939 @@
...
+    return dialect;
+  }
+
+  protected async prepareStatus(role: BlueGreenRole, interimStatus: BlueGreenInterimStatus): Promise<void> {
+    // Detect changes
+    const contextHash: number = this.getContextHash();
+
+    const storedStatus = this.interimStatuses[role.value];
+
+    if (_.isEqual(storedStatus, interimStatus) && this.lastContextHash === contextHash) {
```
aaron-congo, line 161: In JDBC we compare hashes for equality instead of the checking storedStatus and interimStatus directly, do we need to do the same here?

common/lib/plugins/bluegreen/blue_green_status_provider.ts
```diff
@@ -0,0 +1,939 @@
...
+    // Log final switchover results.
+    this.logSwitchoverFinalSummary();
+
+    this.resetContextWhenCompleted();
+  }
+
+  protected updatePhase(role: BlueGreenRole, interimStatus: BlueGreenInterimStatus): void {
+    const latestInterimPhase: BlueGreenPhase = !this.interimStatuses[role.value]
+      ? BlueGreenPhase.NOT_CREATED
+      : this.interimStatuses[role.value].blueGreenPhase;
```
aaron-congo, line 199: Not sure if this would ever actually be a problem in practice but should we store this.interimStatuses[role.value] as a local variable so that we are only accessing it once? I'm wondering if there is a chance that this.interimStatuses[role.value] is not null when we check it the first time, and then becomes null before we call `this.interimStatuses[role.value].blueGreenPhase`, in which case you would be dereferencing a null value
```suggestion
    const latestInterimPhase: BlueGreenPhase = !this.interimStatuses[role.value]
      ? BlueGreenPhase.NOT_CREATED
      : this.interimStatuses[role.value].blueGreenPhase;
```

common/lib/plugins/bluegreen/blue_green_status_provider.ts
```diff
@@ -0,0 +1,939 @@
...
+  protected updateStatusCache(): void {
+    this.pluginService.setStatus<BlueGreenStatus>(BlueGreenStatus, this.summaryStatus, this.bgdId);
+    this.storePhaseTime(this.summaryStatus.currentPhase);
+  }
+
+  protected updateCorrespondingHosts(): void {
+    this.correspondingHosts.clear();
+
+    if (
+      this.interimStatuses[BlueGreenRole.SOURCE.value] &&
```
aaron-congo, line 232: It looks like there are still some fixes from the JDBC bg fix pr that need to be ported over, here and in some other spots. Let me know if you have any questions about them
https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1441

common/lib/plugins/bluegreen/blue_green_status.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  constructor(
+    bgdId: string,
+    phase: BlueGreenPhase,
+    unmodifiableConnectRouting?: ConnectRouting[],
+    unmodifiableExecuteRouting?: ExecuteRouting[],
+    roleByHost?: Map<string, BlueGreenRole>,
+    correspondingHosts?: Map<string, Pair<HostInfo, HostInfo>>
+  ) {
+    this.bgdId = bgdId;
```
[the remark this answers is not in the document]
aaron-congo: Should it be prefixed by an underscore then?

tests/integration/host/build.gradle.kts
```diff
@@ -14,33 +14,37 @@ repositories {
...
+    testImplementation("org.testcontainers:mysql:1.20.4")
+    testImplementation("org.testcontainers:postgresql:1.20.4")
+    testImplementation("org.testcontainers:mariadb:1.20.4")
+    testImplementation("org.testcontainers:junit-jupiter:1.20.4")
+    testImplementation("org.testcontainers:toxiproxy:1.20.4")
+    testImplementation("org.apache.poi:poi-ooxml:5.3.0")
+    testImplementation("org.slf4j:slf4j-simple:2.0.13")
+    testImplementation("com.fasterxml.jackson.core:jackson-databind:2.17.1")
+    testImplementation("com.amazonaws:aws-xray-recorder-sdk-core:2.18.2")
+    testImplementation("io.opentelemetry:opentelemetry-sdk:1.42.1")
```
aaron-congo, line 44: nit: should all the opentelemetry dependencies have the same version?

common/lib/plugins/bluegreen/blue_green_status.ts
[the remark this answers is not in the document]
aaron-congo: Should it be prefixed by an underscore then?

common/lib/plugins/bluegreen/blue_green_status_provider.ts
karenc-bq: _.isEqual is comparing the contents of storedStatus and interimStatus, I believe it achieves the same goal as comparing the custom hashcode.

common/lib/plugins/bluegreen/blue_green_status_provider.ts
```diff
@@ -0,0 +1,936 @@
...
+            this.correspondingHosts.set(blueHostInfo.host, new Pair(blueHostInfo, greenWriterHostInfo));
+          });
+        }
+      }
+    }
+    if (
+      sourceInterimStatus?.startTopology &&
+      sourceInterimStatus?.hostNames?.size > 0 &&
+      targetInterimStatus?.startTopology &&
+      targetInterimStatus?.hostNames?.size > 0
```
aaron-congo, line 260: in JDBC we don't check the startTopology of the source or target interim status, we only check that both statuses are not null and that their host names are not null or empty

common/lib/plugins/bluegreen/blue_green_status_provider.ts
```diff
@@ -0,0 +1,936 @@
...
+      .forEach(([host, role]) => {
+        const hostSpec = this.correspondingHosts.get(host).left;
+        const blueIp = this.hostIpAddresses.get(hostSpec.host);
+        const substituteHostSpecWithIp = !blueIp ? hostSpec : this.hostInfoBuilder.copyFrom(hostSpec).withHost(blueIp).build();
+
+        connectRouting.push(new SubstituteConnectRouting(host, role, substituteHostSpecWithIp, [hostSpec], null));
+
+        connectRouting.push(
+          new SubstituteConnectRouting(
+            this.getHostAndPort(host, this.interimStatuses[role.value].port),
```
aaron-congo, line 503:
```suggestion
        const interimStatus = this.interimStatuses[role.value];
        if (!interimStatus) {
          continue;
        }
        
        connectRouting.push(
          new SubstituteConnectRouting(
            this.getHostAndPort(host, interimStatus.port),
```

tests/integration/container/tests/blue_green_deployment.test.ts
```diff
@@ -0,0 +1,1202 @@
...
+        promiseFinishCount++;
+
+        // Wrapper blue executing connectivity monitoring
+        promises.push(getWrapperBlueExecutingConnectivityMonitoringPromise(hostId, host, instance.port, dbName, stop, results.get(hostId)));
+        promiseCount++;
+        promiseFinishCount++;
+
+        // Wrapper blue new connection monitoring
+        promises.push(getWrapperBlueNewConnectionMonitoringPromise(hostId, host, instance.port, dbName, stop, results.get(hostId)));
+        promiseCount++;
```
aaron-congo, line 375:
```suggestion
        promiseCount++;
        promiseFinishCount++;
```

tests/integration/host/src/test/java/integration/host/TestEnvironmentConfig.java
```diff
@@ -374,6 +583,187 @@ private static void createDbCluster(TestEnvironmentConfig env, int numOfInstance
...
+      if (env.runnerIP == null) {
+        try {
+          env.runnerIP = env.auroraUtil.getPublicIPAddress();
+        } catch (UnknownHostException e) {
+          throw new RuntimeException(e);
+        }
+      }
+      env.auroraUtil.ec2DeauthorizesIP(env.runnerIP);
+      LOGGER.finest(String.format("Test runner IP %s de-authorized. Usage count: %d",
+          env.runnerIP, ipAddressUsageRefCount.get()));
```
aaron-congo, line 754:
```suggestion
      if (!env.reuseDb) {
        env.auroraUtil.ec2DeauthorizesIP(env.runnerIP);
        LOGGER.finest(String.format("Test runner IP %s de-authorized. Usage count: %d",
            env.runnerIP, ipAddressUsageRefCount.get()));
      } else {
        LOGGER.finest("The IP address usage count hit 0, but the REUSE_RDS_DB was set to true, so IP "
            + "de-authorization was skipped.");
      }
```

tests/integration/host/src/test/java/integration/host/TestEnvironmentConfig.java
aaron-congo, line 383:
```suggestion
    env.rdsEndpoint = System.getenv("RDS_ENDPOINT"); // "https://rds-int.amazon.com"
```

aaron-congo (approved):

crystall-bitquill (approved):

PR #528, opened 2025-08-18
fix: parameterized queries

karenc-bq:
### Summary

Add missing API from community drivers.
Update clients to make it more clear which methods are from community drivers and which ones are wrapper-specific


mysql/lib/client.ts
```diff
@@ -221,4 +210,345 @@ export class AwsMySQLClient extends AwsClient {
...
+        },
+        [options, values]
+      );
+    });
+  }
+
+  private async updateState(client: any, options: string | QueryOptions, values?: any): Promise<any> {
+    let sql: string;
+    if (typeof options === "string") {
+      sql = client.format(options, values);
```
sophia-bq, line 542: Might be important to add some sort of null check for client before calling `client.format`?

tests/integration/container/tests/parameterized_queries.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+      const result = await client.query("SELECT $1::int as value, $2::text as name", [42, "test"]);
+      expect(result.rows[0].value).toBe(42);
+      expect(result.rows[0].name).toBe("test");
+    },
+    60000
+  );
+
+  itIfPG(
+    "pg parameterized query with QueryConfig",
```
sophia-bq, line 152: Can there also be a test for QueryArrayConfig?

PR #533, opened 2025-08-29
feat: connection pool

karenc-bq:
### Summary

Update existing interfaces
Introduce AWS Pools
Bug fixes for internal connection pool and error listener

### Description

Updated existing external interfaces to expose more API from community drivers, such as parameterized queries.

Introduced the AwsPGPool and AwsMySQLPool so customers can utilize connection pools from community drivers. These AWS pools implement the **promise**-based APIs from the community drivers, so callbacks are not supported.

Fixed existing bug in the internal connection pool where all pool connections point to the initial connection endpoint, instead of endpoints passed to the create connection calls.

Added code to remove attached error listeners when closing connections.

### DIFF
--- a/pg/lib/client.ts
+++ b/pg/lib/client.ts
@@ -30,7 +30,12 @@ import { DatabaseDialectCodes } from "../../common/lib/database_dialect/database
 import { RdsPgDatabaseDialect } from "./dialect/rds_pg_database_dialect";
 import { PgDatabaseDialect } from "./dialect/pg_database_dialect";
 import { AuroraPgDatabaseDialect } from "./dialect/aurora_pg_database_dialect";
-import { AwsWrapperError, FailoverSuccessError, UnsupportedMethodError } from "../../common/lib/utils/errors";
+import {
+  AwsWrapperError,
+  FailoverSuccessError,
+  UndefinedClientError,
+  UnsupportedMethodError
+} from "../../common/lib/utils/errors";
 import { Messages } from "../../common/lib/utils/messages";
 import { ClientWrapper } from "../../common/lib/client_wrapper";
 import { RdsMultiAZClusterPgDatabaseDialect } from "./dialect/rds_multi_az_pg_database_dialect";
@@ -244,7 +249,7 @@ class BaseAwsPgClient extends AwsClient implements PGClient {
         "query",
         async () => {
           if (!this.targetClient) {
-            throw new AwsWrapperError("targetClient is undefined, this code should not be reachable");
+            throw new UndefinedClientError();
           }
           const sql = typeof queryTextOrConfig === "string" ? queryTextOrConfig : queryTextOrConfig.text;
           await this.pluginService.updateState(sql);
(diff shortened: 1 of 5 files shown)


common/lib/plugin_service.ts
```diff
@@ -332,7 +332,7 @@ export class PluginServiceImpl implements PluginService, HostListProviderService
   }
 
   isBlockingHostListProvider(arg: any): arg is BlockingHostListProvider {
-    return arg;
+    return arg != null && typeof arg.clearAll === "function" && typeof arg.forceMonitoringRefresh === "function";
```
crystall-bitquill, line 335: Should we store things like "function" or "string" as constants somewhere?

pg/lib/client.ts
```diff
@@ -210,4 +190,202 @@ export class AwsPGClient extends AwsClient {
...
+
+  copyTo(queryText: string): Promise<NodeJS.ReadableStream> {
+    return this.pluginManager.execute(
+      this.pluginService.getCurrentHostInfo(),
+      this.properties,
+      "copyTo",
+      async () => {
+        if (!this.targetClient) {
+          throw new AwsWrapperError("targetClient is undefined, this code should not be reachable");
+        }
```
crystall-bitquill, line 277: Do we need this check for the query function?

pg/lib/client.ts
```diff
@@ -210,4 +190,202 @@ export class AwsPGClient extends AwsClient {
...
+    super(config, provider);
+  }
+}
+
+export type { AwsPGPooledConnection };
+
+export class AwsPGPool implements PGPool {
+  private readonly connectionProvider: InternalPooledConnectionProvider;
+  private readonly config;
+  private readonly poolConfig;
```
crystall-bitquill, line 347: Is poolConfig used for anything other than creating the connection provider?

pg/lib/client.ts
```diff
@@ -210,4 +190,202 @@ export class AwsPGClient extends AwsClient {
       null
     );
   }
+
+  // PG Client Implementation
+
+  async connect(): Promise<void> {
+    await this.internalConnect();
+    const context = this.telemetryFactory.openTelemetryContext("AwsClient.connect", TelemetryTraceLevel.TOP_LEVEL);
```
crystall-bitquill, line 198:
```suggestion
    const context = this.telemetryFactory.openTelemetryContext("awsClient.connect", TelemetryTraceLevel.TOP_LEVEL);
```

common/lib/plugin_service.ts
karenc-bq: For this scenario this worked, but if i need to use the value later on, like in this scenario
```
const sql = typeof queryTextOrConfig === ConstantTypes.STRING_TYPE ? queryTextOrConfig : queryTextOrConfig.text;
```
There will be an error 
image
I will just keep it as "string" for now.

common/lib/internal_pooled_connection_provider.ts
```diff
@@ -123,6 +130,9 @@ export class InternalPooledConnectionProvider implements PooledConnectionProvide
   }
 
   async releaseResources() {
+    if (this.internalPool) {
+      await this.internalPool.releaseResources();
```
sophia-bq, line 134: is there a chance this could throw an error?

crystall-bitquill (approved):

PR #610, opened 2026-02-20
refactor: read write splitting

karenc-bq:
### Summary

Refactor read write splitting plugin

### Description

Created an abstract R/W Splitting plugin with common rw splitting functionality

### DIFF
--- a/common/lib/plugin_service.ts
+++ b/common/lib/plugin_service.ts
@@ -150,6 +150,10 @@ export interface PluginService extends ErrorHandler {
   getStatus<T>(clazz: any, key: string): T;
 
   isPluginInUse(plugin: any): boolean;
+
+  isPooledClient(): boolean;
+
+  setIsPooledClient(isPooledClient: boolean): void;
 }
 
 export class PluginServiceImpl implements PluginService, HostListProviderService {
@@ -171,6 +175,8 @@ export class PluginServiceImpl implements PluginService, HostListProviderService
   private allowedAndBlockedHosts: AllowedAndBlockedHosts | null = null;
   protected static readonly statusesExpiringCache: CacheMap<string, any> = new CacheMap();
   protected static readonly DEFAULT_STATUS_CACHE_EXPIRE_NANO: number = 3_600_000_000_000; // 60 minutes
+
+  protected _isPooledClient: boolean = false;
 
   constructor(
     container: PluginServiceManagerContainer,
@@ -781,5 +787,13 @@ export class PluginServiceImpl implements PluginService, HostListProviderService
 
   isPluginInUse(plugin: any) {
     return this.pluginServiceManagerContainer.pluginManager!.isPluginInUse(plugin);
+  }
+
+  isPooledClient(): boolean {
+    return this._isPooledClient;
+  }
+
+  setIsPooledClient(isPooledClient: boolean): void {
+    this._isPooledClient = isPooledClient;
   }
 }
--- a/common/lib/plugins/read_write_splitting/abstract_read_write_splitting_plugin.ts
+++ b/common/lib/plugins/read_write_splitting/abstract_read_write_splitting_plugin.ts
@@ -1,0 +1,311 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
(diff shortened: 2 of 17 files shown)


.github/workflows/main.yml
```diff
@@ -8,6 +8,7 @@ on:
   pull_request:
     branches:
       - "*"
+      - dev/v3
```
karenc-bq, line 10: This is intentional. Will remove when merging dev/v3 branch into main

common/lib/plugins/read_write_splitting/abstract_read_write_splitting_plugin.ts
```diff
@@ -0,0 +1,302 @@
...
+      await this.closeReaderClientIfIdle();
+    }
+
+    logger.debug(Messages.get("ReadWriteSplittingPlugin.switchedFromReaderToWriter", this.writerHostInfo.hostAndPort));
+  }
+
+  async switchToReaderTargetClient() {
+    const currentHost = this.pluginService.getCurrentHostInfo();
+    const currentClient = this.pluginService.getCurrentClient();
+    if (currentHost !== null && currentHost?.role === HostRole.READER && currentClient) {
```
sophia-bq: should we be using some of the helpers here for determining if it is a reader?

common/lib/plugin_service.ts
```diff
@@ -172,6 +176,8 @@ export class PluginServiceImpl implements PluginService, HostListProviderService
   protected static readonly statusesExpiringCache: CacheMap<string, any> = new CacheMap();
   protected static readonly DEFAULT_STATUS_CACHE_EXPIRE_NANO: number = 3_600_000_000_000; // 60 minutes
 
+  protected pooledClient: boolean | null = null;
```
sophia-bq: Wondering if we should have the default value be False and avoid having it be null. From the signature of PluginService and how we use this value we seem to assume it will always be a boolean

common/lib/plugins/read_write_splitting/read_write_splitting_plugin.ts
```diff
@@ -143,251 +81,136 @@ export class ReadWriteSplittingPlugin extends AbstractConnectionPlugin implement
...
-    logger.debug(Messages.get("ReadWriteSplittingPlugin.setWriterClient", writerHostInfo.url));
+    this.writerHostInfo = getWriter(this.hosts, Messages.get("ReadWriteSplittingPlugin.noWriterFound"));
   }
 
-  setReaderClient(readerTargetClient: ClientWrapper | undefined, readerHost: HostInfo): void {
-    this.readerTargetClient = readerTargetClient;
-    this._readerHostInfo = readerHost;
-    logger.debug(Messages.get("ReadWriteSplittingPlugin.setReaderClient", readerHost.url));
-  }
+  protected async forceRefreshAndStoreTopology(currentClient: ClientWrapper | undefined): Promise<void> {
```
sophia-bq: nit: could there be a way to simplify the refresh and forceRefresh methods? they are very similar

sophia-bq (approved):

faviansamatha (approved):

common/lib/plugins/read_write_splitting/read_write_splitting_plugin.ts
```diff
@@ -143,251 +81,136 @@ export class ReadWriteSplittingPlugin extends AbstractConnectionPlugin implement
...
     let targetClient = undefined;
     let readerHost: HostInfo | undefined = undefined;
-    const connectAttempts = this.pluginService.getHosts().length;
+
+    const hostCandidates: HostInfo[] = this.getReaderHostCandidates();
+
+    const connectAttempts = hostCandidates.length * 2;
 
     for (let i = 0; i < connectAttempts; i++) {
       const host = this.pluginService.getHostInfoByStrategy(HostRole.READER, this.readerSelectorStrategy);
```
faviansamatha: Curious, would it make sense to pass in the hostCandidates to getHostInfoByStrategy? If I'm not wrong, default plugin will eventually call pluginService.GetHost which is the same as getReaderHostcandidates, but wondering fi it's worth to just pass it in to keep it consistent to the connectAttemts?

common/lib/plugins/read_write_splitting/read_write_splitting_plugin.ts
karenc-bq: Don't actually need this method, I have removed it

PR #611, opened 2026-02-25
chore: storage service refactoring

karenc-bq:
### Summary

Implement a minimum version of storage service. The full implementation will be implemented along with the full storage service refactoring.

### Description

Implement storage service for Topology class.
Refactor host list providers to use the same topology cache.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -14,6 +14,9 @@ on:
       - "docs/**"
       - "ISSUE_TEMPLATE/**"
       - "**/remove-old-artifacts.yml"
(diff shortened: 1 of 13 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -14,6 +14,9 @@ on:
       - "docs/**"
       - "ISSUE_TEMPLATE/**"
       - "**/remove-old-artifacts.yml"
+  pull_request:
```
karenc-bq, line 17: These workflow changes are intentional. Will remove when merging dev/v3 into main.

PR #615, opened 2026-02-27
chore: topology utils refactoring

karenc-bq:
### Summary

Refactor methods for querying and processing topology information into a TopologyUtils class.


mysql/lib/dialect/aurora_mysql_database_dialect.ts
```diff
@@ -96,6 +112,21 @@ export class AuroraMySQLDatabaseDialect extends MySQLDatabaseDialect implements
     }
   }
 
+  async getInstanceId(targetClient: ClientWrapper): Promise<[string, string]> {
+    const res = await targetClient.query(AuroraMySQLDatabaseDialect.INSTANCE_ID_QUERY);
+    try {
+      const host: string = res[0][0]["instance_id"];
+      const writerId: string = res[0][0]["instance_name"];
```
sophia-bq: Isn't is the instance name? Not the writer id?

mysql/lib/dialect/aurora_mysql_database_dialect.ts
```diff
@@ -41,22 +44,30 @@ export class AuroraMySQLDatabaseDialect extends MySQLDatabaseDialect implements
     "SELECT server_id " +
     "FROM information_schema.replica_host_status " +
     "WHERE SESSION_ID = 'MASTER_SESSION_ID' AND SERVER_ID = @@aurora_server_id";
+  protected static readonly INSTANCE_ID_QUERY: string = "SELECT @@aurora_server_id as instance_id, @@aurora_server_id as as instance_name";
```
sophia-bq:
```suggestion
  protected static readonly INSTANCE_ID_QUERY: string = "SELECT @@aurora_server_id as instance_id, @@aurora_server_id as instance_name";
```

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
```diff
@@ -88,6 +91,7 @@ export class MonitoringRdsHostListProvider extends RdsHostListProvider implement
       this.clusterId,
       () =>
         new ClusterTopologyMonitorImpl(
+          this.topologyUtils,
```
sophia-bq, line 94: I think you also need to pass in the instance template to the monitor

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
```diff
@@ -27,6 +27,7 @@ import { BlockingHostListProvider } from "../host_list_provider";
 import { logger } from "../../../logutils";
 import { SlidingExpirationCacheWithCleanupTask } from "../../utils/sliding_expiration_cache_with_cleanup_task";
 import { isDialectTopologyAware } from "../../utils/utils";
+import { TopologyUtils } from "../topology_utils";
```
sophia-bq, line 30: Are you considering the change where monitors and monitor functionality is added to the rds host list provider so there isn't a monitoring and a rds host list provider?

common/lib/host_list_provider/topology_utils.ts
```diff
@@ -0,0 +1,246 @@
...
+      hosts.push(writers[0]);
+    } else {
+      // Assume the latest updated writer instance is the current writer. Other potential writers will be ignored.
+      const sortedWriters: HostInfo[] = writers.sort((a, b) => {
+        return b.lastUpdateTime - a.lastUpdateTime; // reverse order
+      });
+      hosts.push(sortedWriters[0]);
+    }
+
+    return hosts;
```
sophia-bq, line 227: Based on the way verifyWriter has been written now it will never return an empty list of hosts it will only return null or a list with at least one writer. The line 535 checking for an empty host list in cluster_topology_monitor.ts could probably be simplified

common/lib/host_list_provider/monitoring/monitoring_host_list_provider.ts
karenc-bq: that will be part of the full service container PR

common/lib/host_list_provider/aurora_topology_utils.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { TopologyUtils } from "./topology_utils";
+
+export class RdsTopologyUtils extends TopologyUtils {
```
faviansamatha: I believe the idea of having these topology utils is to have clear separation and roles on dialect and host provider where:
1. dialect would be focused on sql queries that is specific to the db language.
2. topology util: owns executing 
3. host list provider manages the cache of host lists and is only exposed to topology util and not database language specific.


Pros of doing this from what I see is: 
- Removes duplicated parsing logic
- Easier test isolation
- Central point for host construction logic in topology utils
- Dialect doesn't need to care about host construction

pg/lib/dialect/rds_multi_az_pg_database_dialect.ts
```diff
@@ -95,41 +110,18 @@ export class RdsMultiAZClusterPgDatabaseDialect extends PgDatabaseDialect implem
...
       const endpoint: string = row["endpoint"];
       const id: string = row["id"];
       const port: number = row["port"];
       const isWriter: boolean = id === writerHostId;
-
-      const host: HostInfo = hostListProvider.createHost(endpoint.substring(0, endpoint.indexOf(".")), isWriter, 0, Date.now(), port);
-      host.hostId = id;
-      host.addAlias(endpoint);
-      hostMap.set(host.host, host);
+      const host: TopologyQueryResult = new TopologyQueryResult(endpoint.substring(0, endpoint.indexOf(".")), isWriter, 0, Date.now(), port);
```
faviansamatha: Do we need to pass in id? It doesn't look like it's being used here.

common/lib/host_list_provider/topology_utils.ts
```diff
@@ -0,0 +1,249 @@
...
+   */
+  public createHosts(topologyQueryResults: TopologyQueryResult[], initialHost: HostInfo, clusterInstanceTemplate: HostInfo): HostInfo[] {
+    const hostsMap = new Map<string, HostInfo>();
+    topologyQueryResults.forEach((row) => {
+      const hostname = !row.host ? "?" : row.host;
+      const endpoint: string | null = row.endpoint ?? this.getHostEndpoint(hostname, clusterInstanceTemplate);
+
+      const host = this.hostInfoBuilder
+        .withHost(endpoint ?? "")
+        .withPort((row.port ?? clusterInstanceTemplate?.isPortSpecified()) ? clusterInstanceTemplate?.port : initialHost.port)
```
faviansamatha: I think the brackets might be incorrect here.

Just to verify, we want to choose optionA or optionB, whichever isn't null first right? 
optionA = row.Port
optionB = clusterInstanceTemplate port if available, if not get the initial host port? 


I think in the original, we'll treat row.port as a boolean in the ternary if it has a value. 
```suggestion
        .withPort(row.port ?? (clusterInstanceTemplate?.isPortSpecified() ? clusterInstanceTemplate?.port : initialHost.port))
```

common/lib/host_list_provider/aurora_topology_utils.ts
karenc-bq: Discussed this offline, it is worth giving this a thought and try to decouple more of the topology processing logic from database dialects. Will create a ticket for this in the backlog.
This PR will keep using the existing logic where database dialect handles most of the query execution and processing.

sophia-bq (approved):

faviansamatha (approved):

PR #618, opened 2026-03-03
feat: IAM support for GDB

karenc-bq:
### Description

- ensure IAM-based plugins can parse the correct region from global cluster endpoints
- refactor federated auth and okta auth plugins to reduce code duplication


common/lib/utils/iam_auth_utils.ts
```diff
@@ -16,23 +16,26 @@
...
-import { RdsUtils } from "./rds_utils";
 import { Signer } from "@aws-sdk/rds-signer";
 import { AwsCredentialIdentity, AwsCredentialIdentityProvider } from "@smithy/types/dist-types/identity/awsCredentialIdentity";
 import { PluginService } from "../plugin_service";
 import { TelemetryTraceLevel } from "./telemetry/telemetry_trace_level";
+import { HostInfoBuilder } from "../host_info_builder";
 
 export class IamAuthUtils {
-  private static readonly TELEMETRY_FETCH_TOKEN = "fetch IAM token";
+  private readonly TELEMETRY_FETCH_TOKEN = "fetch IAM token";
```
sophia-bq: Why have we made this not static? I can't see any change in use

common/lib/plugins/custom_endpoint/custom_endpoint_plugin.ts
```diff
@@ -106,7 +106,8 @@ export class CustomEndpointPlugin extends AbstractConnectionPlugin implements Ca
       throw new AwsWrapperError(Messages.get("CustomEndpointPlugin.errorParsingEndpointIdentifier", this.customEndpointHostInfo.host));
     }
 
-    this.region = RegionUtils.getRegion(props.get(WrapperProperties.CUSTOM_ENDPOINT_REGION.name), this.customEndpointHostInfo.host);
+    const regionUtils = new RegionUtils();
```
sophia-bq: Curious, is custom endpoint plugin incompatible with global? This doesn't seem to allow it to use the gdb region utils

karenc-bq: Good point, let's add the feature to custom endpoints as well

karenc-bq: Discussed this further offline and decided to keep this unchanged.
We will not reach this section of the code unless the URL is a custom endpoints URL. Custom endpoints URL contains proper region information, so there is no need to retrieve region information using GDBRegionUtils here.

common/lib/utils/gdb_region_utils.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (!clusterId) {
+      return null;
+    }
+
+    const writerClusterArn = await this.findWriterClusterArn(hostInfo, props, clusterId);
+    return writerClusterArn ? this.getRegionFromClusterArn(writerClusterArn) : null;
+  }
+
+  private async findWriterClusterArn(hostInfo: HostInfo, props: Map<string, any>, globalClusterIdentifier: string): Promise<string | null> {
+    if (this.credentialsProvider != null) {
```
faviansamatha:
```suggestion
    if (this.credentialsProvider == null) {
```
I assume this is what u meant? Or else it will always create a new one.

common/lib/utils/gdb_region_utils.ts
```diff
@@ -0,0 +1,102 @@
...
+  private static readonly REGION_GROUP = "region";
+  private credentialsProvider: AwsCredentialIdentity | AwsCredentialIdentityProvider | undefined;
+
+  constructor(credentialsProvider?: AwsCredentialIdentity | AwsCredentialIdentityProvider) {
+    super();
+    this.credentialsProvider = credentialsProvider;
+  }
+
+  async getRegion(regionKey: string, hostInfo?: HostInfo, props?: Map<string, any>): Promise<string | null> {
+    if (props.get(regionKey)) {
```
faviansamatha: Could props be undefined here? I think we might need a null check.

common/lib/utils/gdb_region_utils.ts
```diff
@@ -0,0 +1,102 @@
...
+      return super.getRegion(props.get(regionKey), hostInfo);
```
faviansamatha: Would this fail? region_utils.ts does: 
```
const region = this.getRegionFromRegionString(props.get(regionKey));
```

but you're passing an empty props.

common/lib/utils/region_utils.ts
```diff
@@ -67,21 +68,21 @@ export class RegionUtils {
 
   protected static readonly rdsUtils = new RdsUtils();
 
-  static getRegion(regionString: string, host?: string): string | null {
-    const region = RegionUtils.getRegionFromRegionString(regionString);
+  async getRegion(regionKey: string, hostInfo?: HostInfo, props?: Map<string, any>): Promise<string | null> {
+    const region = this.getRegionFromRegionString(props.get(regionKey));
```
faviansamatha: similar case to above, props could be null

common/lib/utils/region_utils.ts
```diff
@@ -67,21 +68,21 @@ export class RegionUtils {
...
 
     if (region !== null) {
-      return region;
+      return Promise.resolve(region);
```
faviansamatha: Please verify this, but if this is async now, I think it already returns a promise right? I think it's redundant. Same with the other places.

common/lib/plugins/federated_auth/saml_credentials_provider_factory.ts
```diff
@@ -17,15 +17,14 @@
...
-
 import { AwsWrapperError } from "../../utils/errors";
 import { AwsCredentialIdentity, AwsCredentialIdentityProvider } from "@smithy/types/dist-types/identity/awsCredentialIdentity";
 import { decode } from "entities";
 
 export abstract class SamlCredentialsProviderFactory implements CredentialsProviderFactory {
   async getAwsCredentialsProvider(
     host: string,
-    region: string,
+    region: string | null,
```
faviansamatha, line 27: In the  CredentialsProviderFactory, the signature is still:

`getAwsCredentialsProvider(host: string, region: string, ...`

We should change the CredentialsProviderFactory interface.

common/lib/utils/gdb_region_utils.ts
```diff
@@ -0,0 +1,102 @@
...
+*/
+
+import { RegionUtils } from "./region_utils";
+import { HostInfo } from "../host_info";
+import { AwsCredentialsManager } from "../authentication/aws_credentials_manager";
+import { DescribeGlobalClustersCommand, GlobalCluster, GlobalClusterMember, RDSClient } from "@aws-sdk/client-rds";
+import { AwsCredentialIdentity, AwsCredentialIdentityProvider } from "@smithy/types/dist-types/identity/awsCredentialIdentity";
+
+export class GDBRegionUtils extends RegionUtils {
+  private static readonly GDB_CLUSTER_ARN_PATTERN = /^arn:aws:rds:(?<region>[^:\n]*):([^:\n]*):([^:/\n]*[:/])?(.*)$/;
```
faviansamatha: Does GDB exist for gov cloud regions or china? this pattern won't match that since it'll be: 
`arn:aws-us-gov:` and `arn:aws-cn:`

common/lib/utils/gdb_region_utils.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  private async findWriterClusterArn(hostInfo: HostInfo, props: Map<string, any>, globalClusterIdentifier: string): Promise<string | null> {
+    if (this.credentialsProvider != null) {
+      this.credentialsProvider = AwsCredentialsManager.getProvider(hostInfo, props);
+    }
+
+    const rdsClient = this.getRdsClient();
+
+    try {
+      const command = new DescribeGlobalClustersCommand({
```
faviansamatha, line 62: If we get an sdk error e.g. permission issues, will this propagate?

karenc-bq: I will wrap the error but I am leaning towards failing-fast here

common/lib/utils/gdb_region_utils.ts
karenc-bq: We should support these

faviansamatha (approved):

PR #625, opened 2026-03-13
feat: GDB Failover

karenc-bq:


common/lib/host_list_provider/connection_string_host_list_provider.ts
```diff
@@ -112,4 +112,8 @@ export class ConnectionStringHostListProvider implements StaticHostListProvider
   getClusterId(): string {
     throw new AwsWrapperError("ConnectionStringHostListProvider does not support getClusterId.");
   }
+
+  forceMonitoringRefresh(shouldVerifyWriter: boolean, timeoutMs: number): Promise<HostInfo[]> {
+    throw new AwsWrapperError("ConnectionStringHostListProvider does not support forceMonitoringRefresh.");
```
sophia-bq: Can we place this in messages?

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -38,46 +38,47 @@ import { ClientWrapper } from "../../client_wrapper";
...
+  protected readonly failoverReaderHostSelectorStrategy: string = WrapperProperties.FAILOVER_READER_HOST_SELECTOR_STRATEGY.defaultValue;
 
-  constructor(pluginService: PluginService, properties: Map<string, any>, rdsHelper: RdsUtils) {
+  constructor(servicesContainer: FullServicesContainer, properties: Map<string, any>, rdsHelper: RdsUtils) {
     super();
-    this._properties = properties;
-    this.pluginService = pluginService;
-    this._rdsHelper = rdsHelper;
+    this.properties = properties;
+    this.servicesContainer = servicesContainer;
```
sophia-bq, line 75: Do we ever use this.servicesContainer other than to get the pluginService? If not then it doesn't make sense to store it in the plugin

common/lib/wrapper_property.ts
```diff
@@ -210,6 +228,38 @@ export class WrapperProperties {
   );
   static readonly FAILOVER_MODE = new WrapperProperty<string>("failoverMode", "Set host role to follow during failover.", "");
 
+  static readonly FAILOVER_HOME_REGION = new WrapperProperty<string>("failoverHomeRegion", "Set home region for failover.", null);
```
sophia-bq:
```suggestion
  static readonly FAILOVER_HOME_REGION = new WrapperProperty<string>("failoverHomeRegion", "Set home region for GDB failover.", null);
```

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
```diff
@@ -0,0 +1,372 @@
...
+
+    throw new AwsTimeoutError(Messages.get("Failover.failoverReaderTimeout"));
+  }
+
+  protected shortDelay(): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, 100));
+  }
+
+  override async failoverReader(): Promise<void> {
+    throw new UnsupportedMethodError("This method should not be used in this class. See failover() method for implementation details.");
```
sophia-bq: Can we add these to the messages file?

mysql/lib/dialect/global_aurora_mysql_database_dialect.ts
```diff
@@ -77,7 +88,7 @@ export class GlobalAuroraMySQLDatabaseDialect extends AuroraMySQLDatabaseDialect
     rows.forEach((row) => {
       const hostName: string = row["server_id"];
       const isWriter: boolean = row["is_writer"];
-      const hostLag: number = row["visibility_lag_in_msec"] ?? 0; // visibility_lag_in_msec is nullable.
+      const hostLag: number = row["visibility_lag_in_msec"] ?? 0; // visibility_lag_in_sec is nullable.
```
sophia-bq:
```suggestion
      const hostLag: number = row["visibility_lag_in_msec"] ?? 0; // visibility_lag_in_msec is nullable.
```
The value is still called msec?

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
```diff
@@ -0,0 +1,372 @@
...
+            break;
+          default:
+            throw new UnsupportedMethodError(`Unsupported failover mode: ${currentFailoverMode}`);
+        }
+
+        logger.debug(Messages.get("Failover.establishedConnection", this.pluginService.getCurrentHostInfo()?.host ?? "unknown"));
+        this.throwFailoverSuccessException();
+      });
+    } finally {
+      logger.debug(Messages.get("GlobalDbFailoverPlugin.failoverElapsed", String(getTimeInNanos() - failoverStartTimeNs)));
```
sophia-bq: Is the time displayed in nanoseconds or milliseconds? The message mentions ms

common/lib/utils/rds_url_type.ts
```diff
@@ -20,6 +20,7 @@ export class RdsUrlType {
   public static readonly RDS_READER_CLUSTER = new RdsUrlType(true, true, true);
   public static readonly RDS_CUSTOM_CLUSTER = new RdsUrlType(true, false, true);
   public static readonly RDS_PROXY = new RdsUrlType(true, false, true);
+  public static readonly RDS_PROXY_ENDPOINT = new RdsUrlType(true, false, true);
```
sophia-bq, line 23: I think this needs to be added in the same places as RDS_PROXY to be of use to the failover plugins?

tests/unit/stale_dns_helper.test.ts
```diff
@@ -87,208 +79,6 @@ describe("test_stale_dns_helper", () => {
     expect(returnConn).toBe(mockInitialClientWrapper);
   });
 
-  it("test_get_verified_connection_cluster_inet_address_none", async () => {
```
sophia-bq, line 90: Might be good to add some tests for the new functionality of stale dns helper?

tests/integration/container/tests/failover/gdb_failover.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        await auroraTestUtility.simulateTemporaryFailure(initialWriterId!);
+
+        await expect(async () => {
+          await auroraTestUtility.queryInstanceId(client);
+        }).rejects.toThrow(FailoverSuccessError);
+      },
+      1320000
+    );
+  });
+});
```
sophia-bq, line 184: Wondering about adding some tests for when the failover would not be successful? ie. out of region or no allowed hosts?

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
```diff
@@ -0,0 +1,372 @@
...
+        if (!writerCandidate) {
+          this.failoverWriterTriggeredCounter.inc();
+          this.failoverWriterFailedCounter.inc();
+          const message = logTopology(updatedHosts, Messages.get("Failover.unableToDetermineWriter"));
+          logger.error(message);
+          throw new FailoverFailedError(message);
+        }
+
+        // Check writer region to determine failover mode
+        const writerRegion = this.rdsHelper.getRdsRegion(writerCandidate.host);
```
faviansamatha, line 134: Can this return null? Would we have issues if it does?

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
```diff
@@ -0,0 +1,372 @@
...
+    } while (getTimeInNanos() < failoverEndTimeNs); // All hosts failed. Keep trying until we hit the timeout.
+
+    throw new AwsTimeoutError(Messages.get("Failover.failoverReaderTimeout"));
+  }
+
+  protected shortDelay(): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, 100));
+  }
+
+  override async failoverReader(): Promise<void> {
```
faviansamatha: Just to throw a thought out there, would it make sense to just make the failoverReader method private in the base class and not override these to do nothing? 

Answer to that question might be more on what other failover plugins we plan to do.

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
```diff
@@ -0,0 +1,372 @@
...
+            break;
+          case GlobalDbFailoverMode.ANY_READER_OR_WRITER:
+            await this.failoverToAllowedHost(() => [...this.pluginService.getHosts()], null, failoverEndTimeNs);
+            break;
+          default:
+            throw new UnsupportedMethodError(`Unsupported failover mode: ${currentFailoverMode}`);
+        }
+
+        logger.debug(Messages.get("Failover.establishedConnection", this.pluginService.getCurrentHostInfo()?.host ?? "unknown"));
+        this.throwFailoverSuccessException();
```
faviansamatha, line 193: In terms of telemetry, I think we only record it in failoverToWriter. But I don't think we do it for failoverToAllowedHost right?

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
karenc-bq: equalsIgnoreCase has null checks and would return false if writerRegion is null

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
karenc-bq: Makes sense to me

common/lib/utils/storage/storage_service.ts
```diff
@@ -179,18 +200,22 @@ export class StorageServiceImpl implements StorageService {
...
-  get<V>(itemClass: Constructor<V>, key?: unknown): V | null {
+  get<V>(itemClass: Constructor<V>, key?: unknown, registerDataAccess: boolean = true): V | null {
     const cache = this.caches.get(itemClass);
     if (!cache) {
       return null;
     }
 
     const value = cache.get(key);
-    if (value === null || value === undefined) {
+    if (!value) {
```
sophia-bq: Wondering about this change? Example: if the values in the cache were booleans then False would become null

common/lib/plugin_service.ts
```diff
@@ -293,58 +293,46 @@ export class PluginServiceImpl implements PluginService, HostListProviderService
...
-      throw new AwsWrapperError(Messages.get("PluginService.requiredBlockingHostListProvider", typeof hostListProvider));
+
+    if (!this.isDynamicHostListProvider()) {
+      const providerName = hostListProvider?.constructor.name ?? "null";
+      throw new UnsupportedMethodError(Messages.get("PluginService.requiredDynamicHostListProvider", providerName));
     }
 
     try {
-      const updatedHostList: HostInfo[] = await hostListProvider.forceMonitoringRefresh(shouldVerifyWriter, timeoutMs);
+      const updatedHostList: HostInfo[] = await (hostListProvider as any).forceMonitoringRefresh(shouldVerifyWriter, timeoutMs);
```
sophia-bq:
```suggestion
      const updatedHostList: HostInfo[] = await (hostListProvider as DynamicHostListProvider).forceMonitoringRefresh(shouldVerifyWriter, timeoutMs);
```
Why are we casting to any instead of DynamicHostListProvider?

common/lib/utils/messages.ts
```diff
@@ -384,7 +405,23 @@ const MESSAGES: Record<string, string> = {
...
+  "GlobalAuroraTopologyMonitor.cannotFindRegionTemplate": "Cannot find cluster instance template for region '%s'.",
+  "GlobalAuroraTopologyMonitor.invalidTopologyUtils": "TopologyUtils must implement GdbTopologyUtils for GlobalAuroraTopologyMonitor.",
+  "GlobalDbFailoverPlugin.missingHomeRegion":
+    "The 'failoverHomeRegion' property is required when connecting to a Global Aurora Database without a region in the URL.",
+  "GlobalDbFailoverPlugin.missingInitialHost": "Unable to determine the initial connection host.",
+  "GlobalDbFailoverPlugin.startFailover": "Starting Global DB failover procedure.",
+  "GlobalDbFailoverPlugin.isHomeRegion": "Is home region: %s",
+  "GlobalDbFailoverPlugin.currentFailoverMode": "Current Global DB failover mode: %s",
+  "GlobalDbFailoverPlugin.failoverElapsed": "Global DB failover elapsed time: %s ms",
+  "GlobalDbFailoverPlugin.candidateNull": "Candidate host is null for role: %s",
```
sophia-bq: I'm wondering if we could find a better description and name for this error. It's not really capturing the cause which is that we couldn't find a host given the strategy and allowed host list.

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
```diff
@@ -0,0 +1,360 @@
...
+        throw new FailoverFailedError(Messages.get("Failover.unexpectedReaderRole", writerCandidate.host));
+      }
+
+      await this.pluginService.setCurrentClient(writerCandidateConn, writerCandidate);
+      writerCandidateConn = null; // Prevent connection from being closed in finally block
+
+      this.failoverWriterSuccessCounter.inc();
+    } catch (ex) {
+      if (!(ex instanceof FailoverFailedError)) {
+        this.failoverWriterFailedCounter.inc();
```
sophia-bq, line 247: This is a bit confusing could we maybe add a comment explaining that we only want to increment the counter if we haven't already marked it and then thrown a failover failed error? Also what is the benefit to marking it above instead of just always incrementing it here? Is it about timing?

sophia-bq (approved):

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -264,294 +296,521 @@ export class ClusterTopologyMonitorImpl implements ClusterTopologyMonitor {
...
-                this.ignoreNewTopologyRequestsEndTimeMs = 0;
-              } else {
-                this.ignoreNewTopologyRequestsEndTimeMs = Date.now() + this.ignoreTopologyRequestMs;
-              }
-              if (this.highRefreshRateEndTimeMs === -1) {
-                this.highRefreshRateEndTimeMs = 0;
-              } else {
-                this.highRefreshRateEndTimeMs = Date.now() + this.highRefreshPeriodAfterPanicMs;
-              }
+              this.highRefreshRateEndTimeNs = getTimeInNanos() + this.highRefreshRateEndTimeNs;
```
faviansamatha: Should this be 
```suggestion
              this.highRefreshRateEndTimeNs = getTimeInNanos() + BigInt(this.highRefreshRateNs);
```
?

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -264,294 +296,521 @@ export class ClusterTopologyMonitorImpl implements ClusterTopologyMonitor {
...
-                    const hostMonitor = new HostMonitor(this, hostInfo, this.writerHostInfo);
-                    const hostRun = hostMonitor.run();
-                    this.hostMonitors.set(hostInfo.host, hostMonitor);
-                    this.untrackedPromises.push(hostRun);
+              // Update host monitors with the new instances in the topology.
+              const hosts: HostInfo[] | null = this.hostMonitorsLatestTopology;
+              if (hosts && !this.hostMonitorsStop) {
+                hosts.forEach((hostInfo) => {
+                  if (!this.submittedHosts.get(hostInfo.host)) {
+                    const hostMonitor = new HostMonitor(this.servicesContainer, this, hostInfo, this.writerHostInfo);
```
faviansamatha: Curious , do we want to do minimalServiceContainer instead of servicesContainer?

common/lib/utils/utils.ts
```diff
@@ -125,20 +136,39 @@ export function containsHostAndPort(hosts: HostInfo[] | null | undefined, hostAn
   return hosts.some((host) => host.hostAndPort === hostAndPort);
 }
```
faviansamatha, line 132: Btw, I think this method will crash if host is null right? It will be undefined === 0 which is false and then it'll go through host.some()? We should do 

if !hosts || hosts.length === 0) {
return False
}

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -264,294 +296,521 @@ export class ClusterTopologyMonitorImpl implements ClusterTopologyMonitor {
...
-                if (!this.hostMonitors.has(hostInfo.host)) {
-                  const hostMonitor = new HostMonitor(this, hostInfo, this.writerHostInfo);
-                  const hostRun = hostMonitor.run();
-                  this.hostMonitors.set(hostInfo.host, hostMonitor);
-                  this.untrackedPromises.push(hostRun);
-                }
+            await this.closeHostMonitors();
+
+            if (!(hosts !== null && !this.isVerifiedWriterConnection)) {
+              continue;
```
faviansamatha, line 391: I believe this will skip the delay right?

common/lib/host_list_provider/connection_string_host_list_provider.ts
karenc-bq: Updated the interface so this class no longer implements forceMonitoringRefresh

tests/integration/container/tests/failover/gdb_failover.test.ts
karenc-bq: Makes sense, will add it as part of the gdb env integration tests

common/lib/utils/storage/storage_service.ts
karenc-bq: good catch

common/lib/plugins/gdb_failover/global_db_failover_plugin.ts
karenc-bq: Added comment. It is already incremented in Failover2Plugin.

tests/unit/stale_dns_helper.test.ts
karenc-bq: makes sense, will add them in a separate pr

common/lib/plugins/failover2/failover2_plugin.ts
```diff
@@ -464,6 +463,22 @@ export class Failover2Plugin extends AbstractConnectionPlugin implements CanRele
     return methodName === Failover2Plugin.METHOD_END;
   }
 
+  protected initFailoverMode(): void {
+    if (this.rdsUrlType) {
+      return;
+    }
+
+    this.failoverMode = failoverModeFromValue(WrapperProperties.FAILOVER_MODE.get(this.properties));
+    const initialHostInfo: HostInfo = this.hostListProviderService.getInitialConnectionHostInfo();
```
sophia-bq:
```suggestion
    const initialHostInfo: HostInfo = this.hostListProviderService?.getInitialConnectionHostInfo();
```

common/lib/partial_plugin_service.ts
```diff
@@ -0,0 +1,550 @@
...
+  async forceMonitoringRefresh(shouldVerifyWriter: boolean, timeoutMs: number): Promise<boolean> {
+    const hostListProvider = this.getHostListProvider();
+
+    if (!this.isDynamicHostListProvider()) {
+      const providerName = hostListProvider?.constructor.name ?? "null";
+      throw new UnsupportedMethodError(Messages.get("PluginService.requiredDynamicHostListProvider", providerName));
+    }
+
+    try {
+      const updatedHostList = await (hostListProvider as any).forceMonitoringRefresh(shouldVerifyWriter, timeoutMs);
```
sophia-bq: the same fix done in plugin service impl of using dynamic host list provider instead of any should be here

common/lib/host_list_provider/monitoring/cluster_topology_monitor.ts
```diff
@@ -264,294 +296,529 @@ export class ClusterTopologyMonitorImpl implements ClusterTopologyMonitor {
...
+              // Update host monitors with the new instances in the topology.
+              const hosts: HostInfo[] | null = this.hostMonitorsLatestTopology;
+              if (hosts && !this.hostMonitorsStop) {
+                hosts.forEach((hostInfo) => {
+                  if (!this.submittedHosts.get(hostInfo.host)) {
+                    const minimalServiceContainer = ServiceUtils.instance.createMinimalServiceContainerFrom(
+                      this.servicesContainer,
+                      this._monitoringProperties
+                    );
+                    minimalServiceContainer.pluginManager.init();
```
sophia-bq: do we need an await - I believe init of the plugin manager returns a promise?

faviansamatha (approved):

sophia-bq (approved):

PR #626, opened 2026-03-13
feat: GDB RW Splitting

karenc-bq:

### DIFF
--- a/common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin.ts
+++ b/common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin.ts
@@ -1,0 +1,96 @@
+/*
+  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ 
+  Licensed under the Apache License, Version 2.0 (the "License").
+  You may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+ 
+  http://www.apache.org/licenses/LICENSE-2.0
+ 
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+import { ReadWriteSplittingPlugin } from "./read_write_splitting_plugin";
+import { WrapperProperties } from "../../wrapper_property";
+import { HostInfo } from "../../host_info";
+import { RdsUtils } from "../../utils/rds_utils";
+import { ReadWriteSplittingError } from "../../utils/errors";
+import { Messages } from "../../utils/messages";
+import { logger } from "../../../logutils";
+import { ClientWrapper } from "../../client_wrapper";
+import { equalsIgnoreCase } from "../../utils/utils";
+
+export class GdbReadWriteSplittingPlugin extends ReadWriteSplittingPlugin {
+  protected readonly rdsUtils: RdsUtils = new RdsUtils();
+
+  protected restrictWriterToHomeRegion: boolean;
+  protected restrictReaderToHomeRegion: boolean;
+
+  protected isInitialized: boolean = false;
+  protected homeRegion: string;
+
+  protected initSettings(initHostInfo: HostInfo, properties: Map<string, any>): void {
+    if (this.isInitialized) {
+      return;
+    }
+    this.restrictWriterToHomeRegion = WrapperProperties.GDB_RW_RESTRICT_WRITER_TO_HOME_REGION.get(properties);
+    this.restrictReaderToHomeRegion = WrapperProperties.GDB_RW_RESTRICT_READER_TO_HOME_REGION.get(properties);
+
+    this.homeRegion = WrapperProperties.GDB_RW_HOME_REGION.get(properties);
+    if (!this.homeRegion) {
+      const rdsUrlType = this.rdsUtils.identifyRdsType(initHostInfo.host);
+      if (rdsUrlType.hasRegion) {
+        this.homeRegion = this.rdsUtils.getRdsRegion(initHostInfo.host);
+      }
+    }
+
+    if (!this.homeRegion) {
+      throw new ReadWriteSplittingError(Messages.get("GdbReadWriteSplittingPlugin.missingHomeRegion", initHostInfo.host));
+    }
+
+    logger.debug(Messages.get("GdbReadWriteSplittingPlugin.parameterValue", "gdbRwHomeRegion", this.homeRegion));
+
+    this.isInitialized = true;
+  }
+
+  override async connect(
+    hostInfo: HostInfo,
+    props: Map<string, any>,
+    isInitialConnection: boolean,
+    connectFunc: () => Promise<ClientWrapper>
+  ): Promise<ClientWrapper> {
+    this.initSettings(hostInfo, props);
+    return super.connect(hostInfo, props, isInitialConnection, connectFunc);
+  }
+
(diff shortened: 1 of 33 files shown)


common/lib/plugins/read_write_splitting/abstract_read_write_splitting_plugin.ts
```diff
@@ -48,9 +50,10 @@ export abstract class AbstractReadWriteSplittingPlugin extends AbstractConnectio
 
   private _inReadWriteSplit = false;
 
-  protected constructor(pluginService: PluginService, properties: Map<string, any>) {
+  protected constructor(serviceContainer: FullServicesContainer, properties: Map<string, any>) {
     super();
-    this.pluginService = pluginService;
+    this.servicesContainer = serviceContainer;
```
sophia-bq: If we are only using the pluginService we shouldn't need to store the services container

common/lib/utils/messages.ts
```diff
@@ -416,7 +416,14 @@ const MESSAGES: Record<string, string> = {
   "GlobalDbFailoverPlugin.currentFailoverMode": "Current Global DB failover mode: %s",
   "GlobalDbFailoverPlugin.failoverElapsed": "Global DB failover elapsed time: %s ms",
   "GlobalDbFailoverPlugin.candidateNull": "Candidate host is null for role: %s",
-  "GlobalDbFailoverPlugin.unableToConnect": "Unable to establish a connection during Global DB failover."
+  "GlobalDbFailoverPlugin.unableToConnect": "Unable to establish a connection during Global DB failover.",
+  "GlobalDbFailoverPlugin.unableToConnect": "Unable to establish a connection during failover.",
```
sophia-bq: Duplicate message

common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+  protected isInitialized: boolean = false;
+  protected homeRegion: string;
+
+  constructor(serviceContainer: FullServicesContainer, properties: Map<string, any>) {
+    super(serviceContainer, properties);
+    this.restrictWriterToHomeRegion = WrapperProperties.GDB_RW_RESTRICT_WRITER_TO_HOME_REGION.get(properties);
+    this.restrictReaderToHomeRegion = WrapperProperties.GDB_RW_RESTRICT_READER_TO_HOME_REGION.get(properties);
+  }
+
+  protected initSettings(initHostInfo: HostInfo, properties: Map<string, any>): void {
```
sophia-bq, line 36: Curious why some of the properties are gathered in the constructor and then GDB_RW_HOME_REGION is gathered in initSettings? Is it expected that that property value might change as I can't see it modified elsewhere in the code? Or is it to ensure that if needed the error is thrown upon connect?

common/lib/plugins/read_write_splitting/read_write_splitting_plugin_factory.ts
faviansamatha: I believe you'll need to change this or we'll get a runtime error

common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin.ts
```diff
@@ -0,0 +1,102 @@
+/*
```
faviansamatha, line 1: The file name is `gdb_read_writeR_...`, should it just be gdb_read_write_splitting_plugin.ts?

common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+      !equalsIgnoreCase(this.rdsUtils.getRdsRegion(this.writerHostInfo.host), this.homeRegion)
+    ) {
+      throw new ReadWriteSplittingError(
+        Messages.get("GdbReadWriteSplittingPlugin.cantConnectWriterOutOfHomeRegion", writerHostInfo.host, this.homeRegion)
+      );
+    }
+    super.setWriterClient(writerTargetClient, writerHostInfo);
+  }
+
+  protected getReaderHostCandidates(): HostInfo[] {
```
faviansamatha, line 83: This seems a bit off that this returns HostInfo[], but the super's version returns HostInfo | undefined.

common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+      const hostsInRegion: HostInfo[] = this.pluginService
+        .getHosts()
+        .filter((x) => equalsIgnoreCase(this.rdsUtils.getRdsRegion(x.host), this.homeRegion));
+
+      if (hostsInRegion.length === 0) {
+        throw new ReadWriteSplittingError(Messages.get("GdbReadWriteSplittingPlugin.noAvailableReadersInHomeRegion", this.homeRegion));
+      }
+      return hostsInRegion;
+    }
+    return super.getReaderHostCandidates();
```
faviansamatha, line 94: This could be undefined right?

common/lib/plugins/read_write_splitting/gdb_read_writer_splitting_plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+  ): Promise<ClientWrapper> {
+    this.initSettings(hostInfo, props);
+    return super.connect(hostInfo, props, isInitialConnection, connectFunc);
+  }
+
+  override setWriterClient(writerTargetClient: ClientWrapper | undefined, writerHostInfo: HostInfo) {
+    if (
+      this.restrictWriterToHomeRegion &&
+      this.writerHostInfo != null &&
+      !equalsIgnoreCase(this.rdsUtils.getRdsRegion(this.writerHostInfo.host), this.homeRegion)
```
faviansamatha: do we mean to do `this.writerHostInfo`? Or do we mean to do the `writerHostInfo` in the parameter?

common/lib/plugins/read_write_splitting/read_write_splitting_plugin.ts
faviansamatha: Ik not part of this code change, but curious, in node it looks like this is getting cached data, but in JDBC, it looks like it's actually using getHostRole which queries the db.  Curious which one is the expected behaviour.

common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin.ts
karenc-bq: This actually can't return undefined. Updated interface.

common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin.ts
karenc-bq: makes sense to have them all in initsettings

common/lib/plugins/read_write_splitting/gdb_read_write_splitting_plugin_factory.ts
```diff
@@ -0,0 +1,40 @@
...
+ 
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+import { ConnectionPluginFactory } from "../../plugin_factory";
+import { PluginService } from "../../plugin_service";
```
sophia-bq: Think we can remove this import

sophia-bq (approved):

faviansamatha (approved):

PR #646, opened 2026-05-25
chore: update default plugin list

karenc-bq:
### Summary

Add initial connection plugin and failover 2 to the default plugin chain list 
Bump dependencies


common/lib/connection_plugin_chain_builder.ts
```diff
@@ -123,9 +123,12 @@ export class ConnectionPluginChainBuilder {
     } else {
       let pluginCodes: string = props.get(WrapperProperties.PLUGINS.name);
       if (pluginCodes == null) {
-        pluginCodes = WrapperProperties.DEFAULT_PLUGINS;
+        pluginCodes =
+          pluginService.getDriverDialect().getDialectName() === "MySQL2DriverDialect"
+            ? WrapperProperties.MYSQL_DEFAULT_PLUGINS
```
sophia-bq: just a note that however unlikely if we add support for another mysql driver we might need to change the if condition or the naming to WrapperProperties.MYSQL2_DEFAULT_PLUGINS

sophia-bq (approved):

PR #651, opened 2026-05-27
fix: syntax error in integration test yaml

karenc-bq:
### Summary
Fix syntax error in the integration tests workflow


kenrickyap (approved):
LGTM

PR #658, opened 2026-06-16
chore: remove aliases

karenc-bq:
### Summary

- Remove aliases from HostInfo, use identifyConnection to fetch unique instance id instead.
- Update workflow job versions
- Port over fixes for aurora connection tracker plugin, ensure tracked connections are properly drained


common/lib/plugin_service.ts
```diff
@@ -47,7 +47,7 @@ import { AllowedAndBlockedHosts } from "./allowed_and_blocked_hosts";
 import { ConnectionPlugin } from "./connection_plugin";
 import { FullServicesContainer } from "./utils/full_services_container";
 import { StorageService } from "./utils/storage/storage_service";
-import { CoreServicesContainer } from "./utils/core_services_container";
+import type { TrackedConnectionListHost } from "./plugins/connection_tracker/tracked_connection_list";
```
sophia-bq, line 50: I don't think this file exists yet? `"./plugins/connection_tracker/tracked_connection_list"`

common/lib/plugin_service.ts
```diff
@@ -96,14 +96,20 @@ export interface PluginService extends ErrorHandler {
...
-  fillAliases(targetClient: ClientWrapper, hostInfo: HostInfo): Promise<void>;
+  getRoutedHostInfo(): HostInfo | null;
+
+  setRoutedHostInfo(routedHostInfo: HostInfo | null);
 
   identifyConnection(targetClient: ClientWrapper): Promise<HostInfo | null>;
 
+  identifyConnection(targetClient: ClientWrapper, connectionHostInfo: HostInfo | null): Promise<HostInfo | null>;
+
+  identifyConnection(targetClient: ClientWrapper, connectionHostInfo?: HostInfo | null): Promise<HostInfo | null>;
```
sophia-bq, line 111: I don't think any of the plugin service implementations implement these? They only have the original version. The wrapper needs this one for the added value of host_id_cache_service

common/lib/plugins/limitless/limitless_connection_context.ts
```diff
@@ -17,6 +17,7 @@
 import { HostInfo } from "../../host_info";
 import { ClientWrapper } from "../../client_wrapper";
 import { ConnectionPlugin } from "../../connection_plugin";
+import Map from "@arrows/array/src/map";
```
sophia-bq, line 20: What is this?

common/lib/plugin_service.ts
```diff
@@ -162,6 +172,7 @@ export class PluginServiceImpl implements PluginService, HostListProviderService
   private _isInTransaction: boolean = false;
   private servicesContainer: FullServicesContainer;
   protected hosts: HostInfo[] = [];
+  protected routedHostInfo: HostInfo | null;
```
sophia-bq, line 175: Let's set this to null to start - otherwise it is undefined

mysql/lib/client.ts
```diff
@@ -86,7 +87,9 @@ class BaseAwsMySQLClient extends AwsClient implements MySQLClient {
           // Ignore
         }
       }
-      await this.pluginService.setCurrentClient(result, result.hostInfo);
+      const connectedHostInfo: HostInfo | null = this.pluginService.getRoutedHostInfo() ?? this.pluginService.getInitialConnectionHostInfo();
+      await this.pluginService.setCurrentClient(result, connectedHostInfo);
```
sophia-bq, line 91: we might need a null check for host info as I think setCurrentClient needs a non-null host info - same thing in pg client

common/lib/plugins/efm/v1/host_monitoring_connection_plugin.ts
```diff
@@ -110,10 +113,6 @@ export class HostMonitoringConnectionPlugin extends AbstractConnectionPlugin imp
 
         if (monitorContext.isHostUnhealthy) {
           const monitoringHostInfo = await this.getMonitoringHostInfo();
-          if (monitoringHostInfo) {
-            this.pluginService.setAvailability(monitoringHostInfo.allAliases, HostAvailability.NOT_AVAILABLE);
```
sophia-bq, line 114: Was this supposed to be removed?

common/lib/plugins/connection_tracker/opened_connection_tracker.ts
```diff
@@ -21,111 +21,124 @@ import { logger } from "../../../logutils";
...
+    logger.debug(Messages.get("OpenedConnectionTracker.invalidatingConnections", `${host}\n[${connections.join()}\n]`));
   }
 
   pruneNullConnections(): void {
-    for (const [key, queue] of OpenedConnectionTracker.openedConnections) {
-      OpenedConnectionTracker.openedConnections.set(
-        key,
-        queue.filter((connWeakRef: WeakRef<ClientWrapper>) => connWeakRef?.deref() ?? null)
-      );
+    for (const connectionList of OpenedConnectionTracker.openedConnections.values()) {
```
sophia-bq, line 136: should we also remove keys in this list that have an empty queue?

common/lib/plugins/connection_tracker/opened_connection_tracker.ts
```diff
@@ -21,111 +21,124 @@ import { logger } from "../../../logutils";
...
-      .filter((x) => OpenedConnectionTracker.rdsUtils.isRdsInstance(OpenedConnectionTracker.rdsUtils.removePort(x)))
-      .reduce((max, s) => (s > max ? s : max), "");
+    // It might be a custom domain name. Let's track by hostId and custom domain name.
+    let lastHost: TrackedConnectionListHost | null = null;
+    if (hostInfo.hostId) {
+      lastHost = this.trackConnection(hostInfo.hostId, client);
+    }
+    if (hostInfo.hostAndPort) {
+      lastHost = this.trackConnection(hostInfo.hostAndPort, client);
+    }
```
sophia-bq, line 54: If this can add something to be tracked twice, are there any issues if we invalidate the connection twice in invalidateAllConnections?

common/lib/plugin_service.ts
karenc-bq: Thanks, forgot to git add the file

common/lib/plugins/limitless/limitless_connection_context.ts
karenc-bq: Webstorm keeps auto importing classes from this arrows package for me

PR #662, opened 2026-06-22
chore: refactor efm plugins

karenc-bq:
### Summary
- Refactor EFM plugins to use Monitor Service to manage monitors


tests/unit/host_monitoring_plugin.test.ts
```diff
@@ -1,12 +1,12 @@
 /*
   Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
- 
+
```
sophia-bq, line 3: monitor_connection_context.test.ts might need some updates too. could also be a good place to add tests for ContextPoolImpl, HostMonitorImpl, ConnectionContextImpl

common/lib/plugins/efm/v1/host_monitoring_connection_plugin.ts
```diff
@@ -78,40 +83,37 @@ export class HostMonitoringConnectionPlugin extends AbstractConnectionPlugin imp
...
       );
 
-      result = await Promise.race([monitorContext.trackHealthStatus(), methodFunc()])
-        .then((result: any) => {
-          return result;
-        })
-        .catch((error: any) => {
-          throw error;
-        });
+      const raceResult = await Promise.race([this.waitForUnhealthy(context), methodFunc()]);
```
sophia-bq: does race ensure that we cancel or clean up the one that does not finish first? or are we relying on stop monitoring to kill anything eventually?

common/lib/plugins/efm/base/connection_context.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Monitoring context for each connection. This contains each connection's criteria for whether a
+ * host should be considered unhealthy. The context is shared between the main task and the
+ * monitor task.
+ */
+export interface ConnectionContext {
+  readonly failureDetectionIntervalMillis: number;
+  readonly failureDetectionCount: number;
+  readonly expectedActiveMonitoringStartTimeNano: number;
```
sophia-bq, line 31: i think we put nano times as bigint in the other parts of the wrapper

common/lib/plugins/efm/base/host_monitor.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  async releaseResources(): Promise<void> {
+    clearTimeout(this.delayTimeoutId);
+    clearTimeout(this.sleepTimeoutId);
+    await this.stop();
+  }
+
+  async run(): Promise<void> {
+    await super.run();
+  }
```
sophia-bq: i don't think this is necessary - should have the same behavior with this removed

common/lib/plugins/efm/base/host_monitor.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  releaseResources(): Promise<void>;
+}
+
+class ConnectionStatus {
+  constructor(
+    readonly isValid: boolean,
+    readonly elapsedTimeNano: number
+  ) {}
+}
```
sophia-bq: what is the benefit of this vs a simple tuple?

common/lib/plugins/efm/base/context_pool.ts
```diff
@@ -0,0 +1,157 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+import { ConnectionContext } from "./connection_context";
+
+/**
+ * Manages active and pending monitoring contexts grouped by host.
+ */
+export interface ContextPool {
```
sophia-bq: i'm curious if this is as needed for node, as node is single threaded it shouldn't have the same concurrency concerns - could there be a simpler solution for this?

common/lib/plugins/efm/base/connection_context.ts
karenc-bq: The util method `getCurrentTimeNano` returns the current nanos as a Number.
I will create a separate PR going over all the nano variables in the codebase and make it consistent.

common/lib/plugins/efm/base/host_monitor.ts
karenc-bq: I personally thought this may be more readable but I don't really have a strong preference. We can go for a simpler code and switch if complexity increases

common/lib/plugins/efm/v1/host_monitoring_connection_plugin.ts
karenc-bq: waitForUnhealthy should complete on its own eventually and the finally block should clean up methodFunc

common/lib/plugins/efm/base/context_pool.ts
karenc-bq: That's a good point. Although I do wonder if it is even worthwhile introducing the extra complexity. The context pool implementation in JDBC didn't have significant performance impact either.

common/lib/plugins/efm/base/host_monitor.ts
sophia-bq: I'm also okay with keeping it in was just curious!

sophia-bq (approved):

PR #673, opened 2026-07-21
fix: trigger failover on read-only connection errors in strict-writer mode

sudosubin:
### Summary

Trigger failover on read-only connection errors in strict-writer mode, matching the JDBC and Python wrappers.

### Description

After an Aurora writer failover, a pooled connection may still point at the former writer, now demoted to a reader. Writes on it return a read-only error (PostgreSQL SQLSTATE 25006, MySQL errno 1290/1836) rather than a network error. Since `shouldErrorTriggerClientSwitch` only checks `isNetworkError`, failover is never triggered and the connection stays pinned to the old writer. The JDBC and Python wrappers already treat this as a failover trigger in strict-writer mode.

This adds `isReadOnlyConnectionError` to `ErrorHandler` (PG: SQLSTATE 25006; MySQL: errno 1290/1836), delegated through `PluginService` like `isNetworkError`, and makes `Failover2Plugin`/`FailoverPlugin` trigger failover on it in strict-writer mode only (reader-or-writer mode is unchanged). Unit tests included.

### DIFF
--- a/common/lib/error_handler.ts
+++ b/common/lib/error_handler.ts
@@ -23,6 +23,9 @@ export interface ErrorHandler {
 
   isSyntaxError(e: Error): boolean;
 
+  // True when a write was attempted on a read-only connection, e.g. a writer demoted after failover.
+  isReadOnlyConnectionError(e: Error): boolean;
+
   /**
    * Checks whether there has been an unexpected error emitted and if the error is a type of login error.
    */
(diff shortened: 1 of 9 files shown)


karenc-bq (approved):

karenc-bq:
@sudosubin Thank you for checking out this project and contributing this change to the codebase!

PR #678, opened 2026-07-29
chore: add type checks in standard workflow and update node version in actions

karenc-bq:
### Summary
- Add type-checks in main.yml
- Update node 20.x dependency in the workflows to 22.x


kenrickyap (approved):
LGTM

sophia-bq (approved):

PR #679, opened 2026-07-30
chore: gdb regions improvements

karenc-bq:


pg/lib/dialect/pg_database_dialect.ts
```diff
@@ -189,6 +190,10 @@ export class PgDatabaseDialect implements DatabaseDialect {
     return undefined;
   }
 
+  async filterAvailableHosts(hosts: HostInfo[], accessibleRegions: string[]): Promise<HostInfo[]> {
```
sophia-bq: Question: since the method is optional do we even need the implementations for plain mysql and pg dialects? Does it help performance wise to have them? Bc my understanding is that when this is not implemented it has the same behavior as returning all hosts

mysql/lib/dialect/global_aurora_mysql_database_dialect.ts
```diff
@@ -81,6 +85,17 @@ export class GlobalAuroraMySQLDatabaseDialect extends AuroraMySQLDatabaseDialect
     );
   }
 
+  async filterAvailableHosts(hosts: HostInfo[], accessibleRegions: string[]): Promise<HostInfo[]> {
```
sophia-bq, line 86: We only call this once it seems, from the failover plugins, I'm curious why we have the same method defined in two dialects that doesn't seem to be dialect independent - can't we handle this filtering in a shared location?

common/lib/plugins/aurora_initial_connection_strategy_plugin.ts
```diff
@@ -245,6 +252,23 @@ export class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
...
+   * regions are configured, the full host list is returned unchanged. Region filtering is applied
+   * before any strategy or role-based selection so the initial connection never targets an
+   * unreachable region.
+   */
+  private getAccessibleHosts(): HostInfo[] {
+    const hosts = this.pluginService.getAllHosts();
+    if (!this.accessibleRegions) {
+      return hosts;
+    }
+    return hosts.filter((host) => {
```
sophia-bq: I feel like this is similar logic to what is in the dialects - we can probably handle it in the accessible_regions file in utils

common/lib/wrapper_property.ts
```diff
@@ -575,6 +575,33 @@ export class WrapperProperties {
...
+  );
+
+  static readonly GDB_MONITORING_CONNECTION_PRIORITY = new WrapperProperty<string>(
+    "gdbMonitoringConnectionPriority",
+    "Defines the priority for monitoring connections in a Global Aurora Database context. " +
+      "Supports region-aware variants that direct the topology monitor to connect to preferred node types " +
+      "or specific regions. Possible values include: strict-writer-primary, strict-writer-secondary, " +
+      "strict-reader-primary, strict-reader-secondary, writer-or-reader-primary, writer-or-reader-secondary, " +
+      "or a specific AWS region name.",
+    null
```
sophia-bq: Shouldn't the default be strict-writer-primary

pg/lib/dialect/global_aurora_pg_database_dialect.ts
```diff
@@ -90,6 +94,17 @@ export class GlobalAuroraPgDatabaseDialect extends AuroraPgDatabaseDialect imple
...
   }
 
+  async filterAvailableHosts(hosts: HostInfo[], accessibleRegions: string[]): Promise<HostInfo[]> {
+    if (!accessibleRegions || accessibleRegions.length === 0) {
+      return hosts;
+    }
+    const lowerRegions = accessibleRegions.map((r) => r.toLowerCase());
+    return hosts.filter((host) => {
+      const region = this.rdsUtils.getRdsRegion(host.host);
+      return region !== null && lowerRegions.includes(region.toLowerCase());
```
sophia-bq:
```suggestion
      return region !== null && lowerRegions.includes(region);
```
Pretty sure we put it into lower case accesible_regions parsing

common/lib/host_list_provider/monitoring/global_aurora_topology_monitor.ts
```diff
@@ -46,6 +54,48 @@ export class GlobalAuroraTopologyMonitor extends ClusterTopologyMonitorImpl {
 
     this.instanceTemplatesByRegion = instanceTemplatesByRegion;
     this.topologyUtils = topologyUtils;
+    this.accessibleRegions = AccessibleRegions.parse(properties);
+
+    if (this.accessibleRegions) {
+      logger.debug(`GlobalAuroraTopologyMonitor: accessible regions = ${this.accessibleRegions.join(",")}`);
```
sophia-bq: Should this be stored in messages?

sophia-bq (approved):

PR #687, opened 2026-08-17
fix: add configuration types and correct runtime dependencies

karenc-bq:
### Summary

Add TypeScript configuration types for the AWS clients, correct the packaged runtime dependencies, and fix declaration-file resolution.

### Description
**Configuration types**

- Add an `AwsClientConfig` interface (`common/lib/wrapper_property.ts`) covering the wrapper's configuration options, kept in parity with `WrapperProperties`.
- Add driver-specific `AwsMySQLClientConfig` (`extends ConnectionOptions, AwsClientConfig`) and `AwsPgClientConfig` (`extends ClientConfig, AwsClientConfig`), and type the client constructors with them instead of `any`.
- Export `AwsClientConfig`, `AwsMySQLClientConfig`, and `AwsPgClientConfig` from the `./mysql` and `./pg` entry points so consumers can annotate their configs.

**Declaration-file resolution**

- The emitted `.d.ts` files used extensionless relative imports, which are invalid under `node16`/`nodenext` (and silently degrade the types to `any` under `skipLibCheck`). Add a post-build step (`scripts/fix-dts-imports.mjs`, wired into `fix-imports`) that appends `.js`/`/index.js` to relative specifiers in declaration files, mirroring the existing Babel step for `.js` output.

**Runtime / peer dependencies**

- Correct the OpenTelemetry and AWS SDK dependency declarations and add the federated-auth runtime dependencies (`tough-cookie` + `@types/tough-cookie`,  `http-cookie-agent`) so bundlers no longer fail to resolve them.
- Update the Federated Auth and Okta Auth plugin docs to list the required packages accurately.

**Type correctness**

- `SessionStateClient.getTransactionIsolation()` now returns `TransactionIsolationLevel | undefined`, matching `AwsClient` and the implementations (previously surfaced as a `TS2416` for consumers using `skipLibCheck: false`).

**Compatibility notes**

- For TypeScript consumers, `config` is now a closed interface, so inline configuration literals are subject to excess-property checks; the interface is kept in full parity with the supported wrapper properties. JavaScript consumers are unaffected (types are erased; runtime behavior is unchanged).
- Verified via a packed tarball installed into a clean strict/ESM project: the  `./mysql` and `./pg` types resolve and enforce correctly under `bundler`, `node16`, and `nodenext` (with `skipLibCheck` on and off).

### DIFF
--- a/common/lib/wrapper_property.ts
+++ b/common/lib/wrapper_property.ts
@@ -14,8 +14,10 @@
   limitations under the License.
 */
 
+import type { AgentOptions } from "node:https";
 import { ConnectionProvider } from "./connection_provider";
 import { DatabaseDialect } from "./database_dialect/database_dialect";
+import type { AwsCredentialsProviderHandler } from "./authentication/aws_credentials_manager";
 import { AwsWrapperError } from "./utils/errors";
 import { Messages } from "./utils/messages";
 
@@ -75,7 +77,7 @@ export interface AwsClientConfig {
   /** The IAM user used to access the database */
   dbUser?: string;
   /** The options to be passed into the httpsAgent */
-  httpsAgentOptions?: Record<string, any>;
+  httpsAgentOptions?: AgentOptions;
   /** The name or the ARN of the secret to retrieve. */
   secretId?: string;
   /** The region of the secret to retrieve. */
@@ -153,7 +155,7 @@ export interface AwsClientConfig {
   /** A reference to a custom database dialect object. */
   customDatabaseDialect?: DatabaseDialect;
   /** A reference to a custom AwsCredentialsProviderHandler object. */
-  customAwsCredentialProviderHandler?: any;
+  customAwsCredentialProviderHandler?: AwsCredentialsProviderHandler;
   /** Name of the AWS Profile to use for IAM or SecretsManager auth. */
   awsProfile?: string;
   /** Driver configuration profile name */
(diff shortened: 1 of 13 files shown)


pg/lib/client.ts
```diff
@@ -42,6 +42,9 @@ import { isDialectTopologyAware } from "../../common/lib/database_dialect/topolo
 import { PGClient, PGPoolClient } from "./pg_client";
 import { DriverConnectionProvider } from "../../common/lib/driver_connection_provider";
 import { GlobalAuroraPgDatabaseDialect } from "./dialect/global_aurora_pg_database_dialect";
+import { AwsClientConfig } from "../../common/lib/wrapper_property";
+
+export interface AwsPgClientConfig extends ClientConfig, AwsClientConfig {}
```
sophia-bq, line 47: Just wanted to check is there a clean cut of when we use `PG` vs `Pg`?

common/lib/wrapper_property.ts
```diff
@@ -19,6 +19,209 @@ import { DatabaseDialect } from "./database_dialect/database_dialect";
...
+  /** Max number of connection retries the Limitless Connection Plugin will attempt. */
+  limitlessConnectMaxRetries?: number;
+  /** Interval in milliseconds for an Limitless router monitor to be considered inactive and to be disposed. */
+  limitlessTransactionRouterMonitorDisposalTimeMs?: number;
+  /** Map containing any keepAlive properties that the target driver accepts in the client configuration. */
+  wrapperKeepAliveProperties?: Map<string, any>;
+  /** A reference to a custom database dialect object. */
+  customDatabaseDialect?: DatabaseDialect;
+  /** A reference to a custom AwsCredentialsProviderHandler object. */
+  customAwsCredentialProviderHandler?: any;
```
sophia-bq: Just wanted to check if there is any problem with the use of `any` within the wrapper properties? A few of them use `any`

pg/lib/client.ts
karenc-bq: Makes sense to stay aligned. Since existing classes are also inconsistent, I will make this change in a separate PR

sophia-bq (approved):

PR #693, opened 2026-08-20
fix: cluster instance host patterns parsing errors, MySQL IAM authentication and incorrect configuration parameter in documentation

karenc-bq:
### Summary

Fixes four defects that prevented Aurora Global Database support from working as documented: cluster topology was never discovered, the documented `globalClusterInstanceHostPatterns` formats were rejected, MySQL token-based authentication could  not connect, and the Global Database pages named a connection property that does not exist.


docs/using-the-nodejs-wrapper/using-plugins/UsingTheIamAuthenticationPlugin.md
```diff
@@ -39,6 +39,29 @@ The AWS Advanced NodeJS Wrapper supports Amazon AWS Identity and Access Manageme
...
+  plugins: "iam",
+  ssl: {
+    ca: readFileSync("path/to/ssl/certificate.pem").toString()
+  }
+});
+```
+
+If `ssl` is not configured, the wrapper logs a warning and leaves `enableCleartextPlugin` unset, and the connection attempt fails — sending an authentication token unencrypted is not something the wrapper will do implicitly. Set `enableCleartextPlugin: true` yourself only if you have accepted that risk. An explicit value always takes precedence over the wrapper's default.
+
+The same applies to the `federatedAuth` and `okta` plugins, which also authenticate with a generated token. PostgreSQL is unaffected.
```
sophia-bq: Do we want to add a link or a note in the federated auth or okta docs? Or does a link to this section already exist?

sophia-bq (approved):

